PR #3, opened 2021-07-26
Make django_idom an installable app

rmorshea:


Archmonger:
Hey @rmorshea 
I think we might be creating a bit of a misnomer by calling the main IDOM element `def root()`. It feels strange to me to potentially have 20 "roots". Perhaps we can just call it `main`?

rmorshea:
@Archmonger so I came up with a simpler idea, we'll look for a module `<app>.idom` and in that module it must have an `__all__` attribute (this is a [standard attribute](https://docs.python.org/3/tutorial/modules.html#importing-from-a-package) for modules). Anything declared in `__all__` is expected to be a component constructor.

With this `my_app.idom` could look like this:

```python
import idom

__all__ = ["MyComponent"]

@idom.component
def MyComponent():
    ...
```

or it could look like this:

```python
from .my_component import MyComponent
from .my_other_component import MyOtherComponent

__all__ = ["MyComponent", "MyOtherComponent", ...]
```

No need for a standardized naming convention. Just list your component in the `__all__` attribute of your `idom` module.

rmorshea:
Managed to get the `test_app` to run via `python tests/manage.py runserver` but something's wrong in the test suite.

Archmonger:
Utilizing `__all__` isn't very Django.

An alternative could just be a named array. For example, for HTTP views Django uses `urlpattern = [ ]`. We could create a `components` variable and enforce it's declaration in a specific file.

rmorshea:
The problem with `components = [ Demo1, Demo2 ]` is that `components` would need to be declared at the end of the file in order to reference all the defined components, which feels a bit awkward:

```python
@component
def Demo1():
    ...

@component
def Demo2():
    ...

components = [Demo1, Demo2]
```

To avoid this, the components would need to be referenced by name instead of by value. Meaning, it would have to be strings `components = ["Demo1", "Demo2"]`. Given that I thought it would be nice to kill two birds with one stone, by just leveraging `__all__` instead of coming up with a new variable name. After all, if one wanted to define `__all__` in the module you'd end up having to do:

```python
components = ["x", "y", ...]
__all__ = components
```

Additionally, taking this approach does not preclude the possibility of a `myapp/idom/*.py` folder. It would just look like:

```
myapp/
  idom/
    __init__.py
    my_component.py
    my_other_component.py
```

Where `__init__.py` would look like:

```python
from .my_component import MyComponent
from .my_other_component import MyOtherComponent

__all__ = ["MyComponent", "MyOtherComponent"]
```

Let me know your thoughts though.

rmorshea:
Also, just to clarify, taking this approach does not preclude the possibility of a `myapp/idom/*.py` folder. It would just look like this instead:

```
myapp/
  idom/
    __init__.py
    my_component.py
    my_other_component.py
```

Where `__init__.py` would look like:

```python
from .my_component import MyComponent
from .my_other_component import MyOtherComponent

__all__ = ["MyComponent", "MyOtherComponent"]
```

Archmonger:
We wouldn't actually need to define components at the end of the file. If we're following Django design patterns and using [`views.py`](https://github.com/Archmonger/Conreq/blob/main/conreq/core/sign_up/urls.py) and [`urls.py`](https://github.com/Archmonger/Conreq/blob/main/conreq/core/sign_up/urls.py) as an example.

The user could define their IDOM components in any file they want. I personally would argue we can suggest the user reuses `views.py`, or creates a custom `components.py`. Truthfully they could even make a separate folder `/components/demo_1.py`, similar to Django users subfolder their views on larger projects.

Then they `import my component` in all of the components into `idom.py` to follow the same schema as Django's HTTP `urls.py`.

rmorshea:
I'm not sure we'll ever need `idom.py` to do anything other than declare what components are available, so I'm hesitant to require the user to create a whole file that will only every contain:

```python
from .my_component import MyComponent
from .my_other_component import MyOtherComponent
...

components = [MyComponent, MyOtherComponent, ...]
```

Do other projects do this sort of thing?

Archmonger:
That's pretty much what Django's `urls.py` is. The only other project I know of that works outside the framework to "extend" django's core functionality is [Django Channels](https://github.com/django/channels). As you may know, that's extending Django to support websockets. The paradigm they followed is all websockets that have been created need to be imported and placed within one central [`asgi.py`](https://github.com/Archmonger/Conreq/blob/a2a5d1b89fa1f9bb2b4a4cbc561eb9081751690c/conreq/asgi.py).

So I suppose by the notion of Channels, then kind of. But channels doesn't rely on any particular file being within each Django sub-app though.

rmorshea:
I guess you're right, this is pretty much just a different form of `urls.py`. I think you've convinced me, but let me read a bit more on how that works

Archmonger:
I think appropriate filenames for us could potentially be
- `idom.py`
- `components.py`

And for the array could be
- `components = []`
- `parts = []`
- `idom = []`
- `idom_components = []`

Also, we have the option of mandating tuples within the array, if there's some other piece of information you'd like to pass through. For example, `components = [ ('/endpoint/url/', my_component) , ... ]`. I don't think we need this though? Just food for thought.

In the occasion we need more than two pieces of information, we could create a `class component()` and mandate that classtype is within the array.

Archmonger:
Also on that note, Django has a history of forcing the user to create small files with little information in it.

For example, check out [`apps.py`](https://github.com/Archmonger/Conreq/blob/main/conreq/core/sign_up/apps.py).

Django is from 2005 which hasn't changed it's design syntax since. Some syntax can feel weird due to that.

rmorshea:
I like `idom.py` as a file name - `components.py` is generic enough that it could clash with something in an existing project. Given that, the  name `idom_components` feels repetitive since it'll already be in a file called `idom`. With all that in mind, I think requiring a `components` attribute to be present in `myapp.idom` seems appropriate. I'll make those changes now.

Archmonger:
I just had an idea while thinking about Django Channels.

We could mandate the user to centralize all components into one file similar to how Channels does it.

Basically, in `settings.py` have a setting called `IDOM_COMPONENTS`, which is an importable path to the IDOM config file.

For example: `IDOM_COMPONENTS = "conreq.core.idom"`

Then `idom.py` file will have one giant `components = []`, and we mandate the user to import all components into this one file.

rmorshea:
So basically, the file where the components need to be declared would be configurable?

Archmonger:
Yeah, configurable but it's all centralized to one file.

rmorshea:
Ah, I see. I think that might break things since in the template you have to reference the component via `idom_view "my_app.MyComponent"`

rmorshea:
Checking each app for an `idom` module isn't that bad anyway given that the list of apps is always available via `INSTALLED_APPS`.

Archmonger:
Thought about it, but we should be able to get around that.

We can inspect each individual function's relative import path and utilize that as a unique identifier. Would be globally unique for certain.
```python
import inspect
inspect.getmodule(my_component).__name__
'conreq.apps.sign_up.idom.my_component'
```

rmorshea:
Headed off to bed, will revisit this tomorrow.

rmorshea:
@Archmonger I think this is good for a final review before merging.

Archmonger:
@rmorshea I accepted the team invite. I'll go ahead and review tomorrow if you assign me. 

On the note of those two function names, would you be willing to consider a less-wordy rename?

README.md
```diff
@@ -10,43 +10,169 @@
   <img alt="License: MIT" src="https://img.shields.io/badge/License-MIT-purple.svg">
 </a>
 
-A package for building highly interactive user interfaces in pure Python inspired by
-[ReactJS](https://reactjs.org/).
+`django-idom` allows you to integrate [IDOM](https://github.com/idom-team/idom), a
+package for building highly interactive user interfaces in pure Python inspired by
```
Archmonger, line 14: Run on sentence. Consider shortening. Readers typically have a short attention span.

rmorshea:
Ok, should be good now.

Archmonger:
Did not have time to complete my review today. Will make time for it this weekend.

src/django_idom/templatetags/idom.py
```diff
@@ -0,0 +1,38 @@
...
+register = template.Library()
+
+
+# Template tag that renders the IDOM scripts
+@register.inclusion_tag(TEMPLATE_FILE_PATHS["head_content"])
+def idom_head():
+    pass
+
+
+@register.inclusion_tag(TEMPLATE_FILE_PATHS["view"])
```
Archmonger, line 25: Looks like we're providing `register.inclusion_tag` a full file path to the bare _development_ template file. This will result in unusable code for production environments where `DEBUG` is turned off and the user uses `python manage.py collectstatic`.

You should only pass in a sub-path paths that of what comes after the `templates/` dir.

To mimic this we need to `lstrip()` anything before `idom/templates/` (inclusive).

Example:
Original path: `/home/my_app/idom/templates/idom/views.html`
New path: `idom/views.html`

src/django_idom/websocket_consumer.py
```diff
@@ -1,20 +1,24 @@
...
 
 from channels.generic.websocket import AsyncJsonWebsocketConsumer
 from idom.core.dispatcher import dispatch_single_view
 from idom.core.layout import Layout, LayoutEvent
-from idom.core.proto import ComponentConstructor
+
+from .app_components import get_component, has_component
+
+
+logger = logging.getLogger(__name__)
```
Archmonger, line 15: Make logger private (`_logger`). Follow this convention across other files as well.

Archmonger (changes requested):

Archmonger:
@rmorshea Overall the design pattern is solid.

- Handful of "cleanup" suggestions.
- Two major issues in relation to `register.inclusion_tag`.

README.md
```diff
@@ -10,43 +10,220 @@
...
+
+This file is automatically discovered by `django-idom` when scanning the list of
+[`INSTALLED_APPS`](https://docs.djangoproject.com/en/3.2/ref/settings/#std:setting-INSTALLED_APPS).
+All apps that export components will contain this module.
+
+Inside this module must be a `components` list that is imported from
+[`components.py`](#components.py):
+
+```python
+from .components import Hello
```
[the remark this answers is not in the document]
rmorshea: This convention actually comes from React, and is a style that I'm going to try and encourage since it will allow this [Flake8 Plugin](https://github.com/idom-team/flake8-idom) to enforce the "rules of hooks".

Will add a second component though to better demo the list.

src/django_idom/templatetags/idom.py
rmorshea: For some reason I think I was having issues with using the string in CI. Will try and revert back to using a string instead to be sure.

Archmonger: If you're having issues with sub paths then it's possible your template related settings in settings.py are misconfigured.

rmorshea:
my computer is acting up - so it'll be a bit before I can get to this

README.md
```diff
@@ -10,43 +10,220 @@
...
 pip install django-idom
 ```
 
 # Django Integration
 
-This version of IDOM can be directly integrated into Django. For example
+To integrate IDOM into your application you'll need to modify or add the following files to `your_app`:
+
+```
+your_app/
```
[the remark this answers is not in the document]
rmorshea: I updated the structure in the README. Let me know if that needs more work.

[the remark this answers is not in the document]
Archmonger: Apps created by Django's default `manage.py startapp <app_name>` command have the following files: `admin.py`, `apps.py`, `models.py`, `tests.py`, and `views.py`.

Although `urls.py` is a extremely common convention it's not a default.

In this example I would put in all of Django's default files, remove `urls.py` (since it's technically not a default), and then add in your two custom app files.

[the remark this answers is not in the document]
rmorshea: I think adding files that aren't discussed below might be confusing. I really wanted to just highlight what was important rather than show what all the files in a working app might be.

README.md
```diff
@@ -10,43 +10,220 @@
...
+```html
+<!-- don't forget your load statements -->
+{% load static %}
+{% load idom %}
+
+<!DOCTYPE html>
+<html>
+  <body>
+    ...
+    {% idom_view "test_app.Hello" name="World" %}
```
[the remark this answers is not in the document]
rmorshea: The author should have access to the component name when they declare it, so I think they'd be able to set that default themselves - statically or, with something like `partial`.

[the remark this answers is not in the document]
rmorshea: And just to clarify `name` refers to a parameter of the `Hello` component.

README.md
```diff
@@ -10,43 +10,220 @@
...
-# Example code goes here
+# the base URL for all IDOM-releated resources
+IDOM_BASE_URL: str = "_idom/"
+
+# ignore these INSTALLED_APPS during component collection
+IDOM_IGNORE_INSTALLED_APPS: set[str] = {"some_app", "some_other_app"}
 ```
 
-For examples on how to use IDOM, [read the IDOM Documentation](https://idom-docs.herokuapp.com).
+# Developer Guide
```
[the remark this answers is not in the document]
rmorshea: At some point I'd like to make Sphinx-based docs for this. The `README.md` should ultimately just be good enough to get by on for a while.

README.md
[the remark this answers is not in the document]
Archmonger: Ahh I understand. If name is a parameter to the hello component perhaps a comment is needed above this in order to explain that.

README.md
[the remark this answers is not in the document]
Archmonger: Sounds good. I still recommend changing `sub_app/` to `example_app/` or `my_app/`. Especially in this example, as the project root isn't an app. 

In `django_idom`'s repository structure the project root is both an app and the root, so `sub_app/` would be appropriate in that case. But that doesn't hold true to the example project structure in this readme.

README.md
```diff
@@ -10,43 +10,261 @@
...
+to render (of the form `app_name.ComponentName`) and keyword arguments you'd like to
+pass it from the template.
+
+```python
+idom_view app_name.ComponentName param_1="something" param_2="something-else"
+```
+
+In context this will look a bit like the following...
+
+```html
```
Archmonger, line 183: change this to markdown codeblock from `html` to `jinja` for better syntax highlighting

README.md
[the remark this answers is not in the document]
rmorshea: I might change the parameter name too - `name` is rather generic and easily confused.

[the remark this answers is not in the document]
Archmonger: Agreed.

Perhaps `example_param` would do the trick?

Archmonger:
How do you feel about supporting a decorator (or an API to manually register a component) alongside the standard `idom.py` paradigm?

_Disclosure:_ Am peddling this because although Django doesn't utilize decorators, I won't be following Django design patterns for Conreq community apps. Decorators will be used to extend Conreq similar to flask. So ideally I'd like to keep that design pattern for Conreq community app developers using `django-idom`.

rmorshea:
I like the idea of decorator approach (since that's what I'm personally more comfortable with), but I think we should choose one way or the other. Having two ways to do the same thing can be confusing for users. I'll defer to you on which we should use though.

Here are some ideas for decorator APIs:

```python
from django_idom import Components

components = Components(
    # auto-infer the module __name__ by default
    module_name=None,
    # whether to auto-apply the idom.component decorator
    is_render_function=True,
)

@components.register
def MyComponent():
    ...
```

```python
from django_idom import register_component

@register_component(
    # auto-infer the module __name__ by default
    module_name=None,
    # whether to auto-apply the idom.component decorator
    is_render_function=True,
)
def MyComponent():
    ...
````

**Comment by rmorshea** [comment]
Even if we go with the auto-collection approach, a decorator in Conreq could be made to work. All you'd have to do is implement a decorator that would add components to the list in Conreq's `idom.py` file. We could make some slight changes to the auto-collection mechanism to make component naming in that scenario a little easier too.

**Comment by rmorshea** [comment]
If we went with the decorator approach, I don't think you'd even need to do anything special with Conreq. Once a user registered the component using `django_idom` the component would be globally accessible via its `module_name.ComponentName` in templates.

**Comment by Archmonger** [comment]
Ideally I'd like to maintain Django design patterns for the sake of consistency (make everything "feel like Django"). But the issue is Django design patterns aren't as clean as more modern frameworks.

After thinking about it, I think it's fine to violate the design patterns for the sake of modernization.

I'm voting for the `register_component` syntax.

**Comment by rmorshea** [comment]
So a subtle issue I came across while implementing the decorator is that because we're not auto-importing the module where components get defined anymore, if the user doesn't ensure that module is imported before Django loads the templates, they'll get an error saying that `django_idom` doesn't know about their components (since they haven't been registered yet).

We could...

1. Tell the user to add `import some_app.components` to their `views.py`
2. Auto-import `components.py` from each app and tell them to register their components in that file.
3. Have them explicitly declare where components are registered (similar to Django's URL `include` function).

The latter two definitely fall back into standard Django-esk patterns.

**Comment by rmorshea** [comment]
So `django_idom.utils.register_components()` doesn't really solve the problem since, if you're able to call it in your components module, then it's already been imported and registration would have happened anyway - the `register_component` decorator works fine once the modules where it's used get imported.

The problem which needs to be solved is that the components module must be imported before the app starts to load its templates (where we reference registered components by name).

**Comment by Archmonger** [comment]
You're right. Hmm... I say we just auto discover components.py with `autodiscover_modules()`.

**Comment by rmorshea** [comment]
Ah, I just had an idea. Given that you have to reference the components in your templates, I can just load them based on that name. Don't even need a `register` decorator anymore either!

```jinja
<body>
    ...
    {% idom_view "your_app.sub_app.components.Hello" name="World" %}
</body>
```

Archmonger:
You think we'd need a schema to cache imports in that scenario?

rmorshea:
I can add a check in `sys.modules` before using `importlib`, and yeah, it's performance isn't great, but I'm sure it would be fine in this situation since it'd only impact the time-to-start.

rmorshea:
This has the additional advantage that the user won't have to "re-export" components that come from other packages. Where in prior loading methods the user would have to do something like:

```python
from some_other_package import SomeComponent

components = [SomeComponent, ...]
```

The user can just reference `SomeComponent` directory from the package in their template without any extra work.

```jinja
<body>
  ...
  {% idom_view "some_other_package.SomeComponent" %}
</body>
```

Archmonger:
Since the module loading is dynamic based on template load, I would think it would impact template first render speed more so than time-to-start?

rmorshea:
I added a check for already imported modules in `sys.modules` so we won't call `import_module` more than we need to.

rmorshea:
Beyond that, unless you want to do a final review pass. I think this is finally ready.

Archmonger:
Sure, I'll do a quick pass on it by this weekend.

src/django_idom/templatetags/idom.py
```diff
@@ -0,0 +1,54 @@
...
+    IDOM_WEB_MODULES_URL,
+    IDOM_WEBSOCKET_URL,
+)
+
+
+register = template.Library()
+
+
+@register.inclusion_tag("idom/view.html")
+def idom_view(_component_id_, **kwargs):
```
Archmonger, line 20: Do we want to consider renaming this to `idom_component`?

Syntactically it makes more sense to me for a "Django view" to contain an embedded "idom component", rather than another "view".

This change would make a correlation to the fact that we're loading by component via import.
ex) `{{ idom_component <IMPORT_PATH> }}`

Archmonger (changes requested):
Nine minor comments in relation to syntax, comment, or verbiage. 

Overall fantastic, developer friendly, and well thought out design.

**Very excited for this!**

README.md
```diff
@@ -10,43 +10,230 @@
...
+http_asgi_app = get_asgi_application()
+
+from channels.routing import ProtocolTypeRouter, URLRouter
+
+application = ProtocolTypeRouter(
+    {
+        "http": http_asgi_app,
+        "websocket": URLRouter(
+          # add a path for IDOM's websocket
+          [idom_websocket_path()]
```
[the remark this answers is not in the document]
rmorshea: I pass on kwargs to `django.urls.path` right now but, if that's not really useful, then yes, this could be a constant.

README.md
```diff
@@ -10,43 +10,230 @@
...
+
+You'll need to include IDOM's static web modules path using `idom_web_modules_path`.
+Similarly to the `idom_websocket_path()`. If you wish to change the route where this
+websocket is served from, see the available [settings](#settings.py).
+
+```python
+from django_idom import idom_web_modules_path
+
+urlpatterns = [
+    idom_web_modules_path(),
```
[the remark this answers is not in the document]
rmorshea: see above.

noxfile.py
```diff
@@ -51,9 +53,16 @@ def test_suite(session: Session) -> None:
 
     session.chdir(HERE / "tests")
     session.env["IDOM_DEBUG_MODE"] = "1"
-    session.env["SELENIUM_HEADLESS"] = str(int("--headless" in session.posargs))
-    session.run("python", "manage.py", "build_js")
```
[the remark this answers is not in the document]
rmorshea: The JS gets rebuilt whenever you run `pip install -e .`, so we can probably just direct people to do that for now.

If we were to have a dedicated command for this, we could put it in the `noxfile.py`. In the main `idom` it's worked out great to be able to tell users that all the "scripty" commands they need can be run with `nox`.

README.md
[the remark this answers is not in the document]
Archmonger: I don't think the user has any need to ever pass in kwargs. That's really only used to pass arguments into the initialization of a consumer (or view), and we don't have any.

If someone was writing their own consumer that does accept params, they would have had to write their own custom path anyways.

So I think we're safe to have `IDOM_WEBSOCKET` and `IDOM_WEB_MODULES` vars.

src/django_idom/views.py
```diff
@@ -0,0 +1,7 @@
+from django.http import HttpRequest, HttpResponse
+from idom.config import IDOM_WED_MODULES_DIR
+
+
+def web_modules_file(request: HttpRequest, file: str) -> HttpResponse:
+    file_path = IDOM_WED_MODULES_DIR.current.joinpath(*file.split("/"))
+    return HttpResponse(file_path.read_text(), content_type="text/javascript")
```
Archmonger, line 7: Rendering javascript dynamically is frowned upon in Django for performance reasons. Plus we're hitting the filesystem on every function call, which is horrible for IOPS.

Can I get a TLDR why we need these web modules to be dynamic?

rmorshea: There might be a way to do this since `IDOM_WED_MODULES_DIR` can be changed. However files get added to the `IDOM_WED_MODULES_DIR` dynamically at runtime, so we'll need to know the directory where the static JS files will ultimately be served from.

rmorshea: It actually looks like this might be another thing we could manage for the user by automatically placing the dynamically added files within `STATIC_ROOT`. I might not be understanding how [`collectstatic`](https://docs.djangoproject.com/en/3.2/ref/contrib/staticfiles/#django-admin-collectstatic) works though.

Archmonger: In order for `collectstatic` to work, the files need to be present during Django's boot up sequence.

I still don't understand what web modules are, but can we compile them during build time? If so, then Django's staticfile system would be able to serve them.

rmorshea: For context, "web module" is't really a technical term, it's just what I've been calling Javascript files that expose an interface IDOM expects. The `IDOM_WEB_MODULES_DIR` then, is the directory where these bundles get collected at runtime. Thus, the full set of files within `IDOM_WEB_MODULES_DIR` are not known at build-time.

With that said, the location of the directory can be changed. So if we know where within `STATIC_ROOT` we should drop these JS files such that they won't clash with those compiled from `collectstatic`, then I think we can make this work. In a sense, IDOM has a similar concept of static file collection, but one which happens later than Django's.

Archmonger: You absolutely don't want to ever manually place files into `STATIC_ROOT`. That would break schemas such as the collectstatic [manifest backend](https://docs.djangoproject.com/en/3.2/ref/contrib/staticfiles/#django.contrib.staticfiles.storage.ManifestStaticFilesStorage). Additionally, it would be completely incompatible with python static file servers such as [whitenoise](https://github.com/evansd/whitenoise).

Does this **need** to be done at runtime? Is it possible to somehow have a command within IDOM to pre-compile these components at build time?

rmorshea: There are several ways to load JS components in IDOM:

1. [dynamically](https://idom-docs.herokuapp.com/docs/javascript-components.html#dynamically-loaded-components)
2. [statically via PyPI](https://idom-docs.herokuapp.com/docs/javascript-components.html#distributing-javascript-via-pypi)
3. [from some url](https://idom-docs.herokuapp.com/docs/javascript-components.html#distributing-javascript-via-cdn) (usually a CDN)

Static file collection isn't possibly for (1), is possible for (2) but would require breaking changes in IDOM, and isn't needed for (3).

I'm not sure there's a great work around for this at the moment besides recommending that people who need performance use method (3) and serve their JS from a CDN.

Archmonger: ### My Thoughts
- **Option 1**
   - I'm assuming this the currently developed solution within `django-idom`.
   - Not ideal for this repo, causes large performance impact and increases IOPS unnecessarily.
   - If we set the `IDOM_WED_MODULES_DIR` to `django-idom/static/js`, is there any way we can get it to build early?
      - Does the IDOM repo have a function handler for creating web modules?
      - If so then we could potentially call this in `django_idom`'s `__init__` which is guaranteed to occur before collecstatic happens.
- **Option 2 (Preferred)**
   - This is ideal from a design perspective, but if there isn't any easy workaround for this then I recommend we just open an issue ticket for `django-idom` and handle it later.
- **Option 3**
   - Some Django users really like self hosting all their own files, but prepackaging JS on a CDN isn't the worst option. Would require us to reintroduce the `{% idom_scripts %} template tag.

rmorshea: I think we should organize a quick chat to better explain all the working pieces here.

We might be able to use the templates and the `idom_component` declarations to find all the modules where web modules are defined. There'd still be some edge cases potentially, but pretty unlikely.

Archmonger: I'm available today, and after 5 tomorrow.

rmorshea: I won't be available this weekend or the next. Monday would work though.

Archmonger: I'll be available Monday after 5, does that work?

rmorshea: Will make a calendar event when I find time.

src/django_idom/views.py
```diff
@@ -0,0 +1,7 @@
+from django.http import HttpRequest, HttpResponse
+from idom.config import IDOM_WED_MODULES_DIR
+
+
+def web_modules_file(request: HttpRequest, file: str) -> HttpResponse:
```
rmorshea, line 5: Add `lru_cache` decorator here.

Archmonger: generic `funcutils.cache` decorator would work fine here too, since we don't really want to ever limit the amount of modules being cached.

src/django_idom/views.py
rmorshea: This conversation is now being tracked here: https://github.com/idom-team/django-idom/issues/4

README.md
```diff
@@ -194,7 +194,7 @@ def your_view(request):
 
 ## `example_app/urls.py`
 
-Include your replate in the list of urlpatterns
+Include your template in the list of urlpatterns
```
Archmonger, line 197: Replace with "your template view" or "your view"

README.md
```diff
@@ -10,43 +10,243 @@
...
 ```bash
 pip install django-idom
 ```
 
 # Django Integration
 
-This version of IDOM can be directly integrated into Django. For example
+To integrate IDOM into your application you'll need to modify or add the following files to `your_app`:
+
+```
```
Archmonger, line 39: Rename `your_app` to `your_project`, `django_project`, or `example_project` since this top level folder in this example isn't an installable app.

Archmonger (approved):
Looks good!

Archmonger (approved):

rmorshea:
Weird, your review doesn't satisfy the merge requirements. Lemme see what's going on there...

Archmonger:
It's because I don't have write access to the repo.

rmorshea:
I'll give that to you.

rmorshea:
And we're all green finally!

rmorshea:
I'll make a release later this evening.

rmorshea:
Released!

Archmonger:
👍🍻

I'll begin Conreq integration soon. 

Would you have time to discuss Conreq design decisions this weekend? I want to make sure it's well aligned with IDOM for interoperability.

rmorshea:
I'll be away from my computer, but I may be able to just voice chat. I'd be able to do a normal video chat Friday afternoon or next week.

Archmonger:
Friday afternoon works with me. I don't have anything scheduled for Friday so I'll add something on your calendar.

PR #11, opened 2021-09-09
synchronize JS and Py version requirements for IDOM

rmorshea:
The IDOM client and core libraries now have synchronized versions. To ensure they work correctly together we need to ensure the version requirements we use both are the same.

### DIFF
--- a/requirements/pkg-deps.txt
+++ b/requirements/pkg-deps.txt
@@ -1,2 +1,2 @@
 channels<4.0.0 # Django websocket features
-idom<1.0.0 # Python React
+idom >=0.33.0, <0.34.0
--- a/src/js/package.json
+++ b/src/js/package.json
@@ -18,6 +18,6 @@
     "rollup-plugin-replace": "^2.2.0"
   },
   "dependencies": {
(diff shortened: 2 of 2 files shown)


Archmonger:
Are we going to configure an automated method of updating the JS and Py versions, or will it have to manually be bumped on each django-idom version release?

rmorshea:
For now I've just been manually bumping the version for `django-idom`, `idom-jupyter`, etc.

Figuring out an automated way to do that would be nice though. Ideally, we could just have a chron job in GitHub actions that checks to see if there's a new version of IDOM available and then automatically creates a PR with the relevant changes if there is.

rmorshea:
Actually it might be easier to generate an issue for each downstream repo when a new release of IDOM is published. Not quite as automated, but it's also less infra to maintain.

Archmonger:
@rmorshea 
Since current pypi installs can be potentially out of sync due to unpinned versions, I think it's a good idea to push out a pypi update to resolve this.

rmorshea:
I can't tonight, but I'll create a quick PR tomorrow to bump the version and then make a release.

Issue #4, opened 2021-08-17
Improve performance for accessing files from IDOM_WEB_MODULES_DIR


rmorshea:
`IDOM_WEB_MODULES_DIR` contains JS modules that are dynamically added at runtime. Since JS files are typically loaded statically, it would be nice if we could improve how quickly these files can be accessed. There are three ways to do that:

- [ ] Implement a management command for collecting these static files ahead of time.
- [x] Make the view accessing the files `async` to avoid blocking while waiting on the file system
- [x] Cache this view using [Django's cache framework](https://docs.djangoproject.com/en/3.2/topics/cache/).

### Notes

To implement the management command we could trace over all the templates and look for `idom_component` template tag nodes in order to figure out what modules to import so that we can trigger the collection of the JS files into `IDOM_WEB_MODULES_DIR`. Then once they've been collected those files can be copied into a permanent static location.

To do this we can take inspiration from the [django compressor](https://github.com/django-compressor/django-compressor) library which takes a similar approach. See the `COMPRESS_OFFLINE` setting to figure out how they're doing it. Also the `compressor.offiline` module contains uniform interfaces for parsing template syntax for Django and Jinja that we might want to copy (with the latest [LICENSE](https://github.com/django-compressor/django-compressor/blob/develop/LICENSE), what to do about `AUTHORS` file?).

---

_Conversation originated from: [this comment](https://github.com/idom-team/django-idom/pull/3/files/4bcdeb522ad86d46ad7e3bee13ad51bdc32026fc#r688325126)_

rmorshea:
@Archmonger I'm looking at what's required to implement this and its kind of ugly stuff. I'd mostly be copy-pasting code I don't really understand which makes me nervous about maintaining it going forward. I'm also still a bit skeptical that this will actually be a bottle neck. Given that, I think I'm going to hold off for now. If it does prove to be a real issue, we can revisit at that point.

With that said, I think there are two simpler solutions to this problem of the dynamic files in `IDOM_WEB_MODULES_DIR` that can improve performance and will be easier to maintain:

1. Make the view reading these web modules `async` so it can `await` the file instead of blocking until the file system is available.
2. Allow users to configure a cross-process cache using [Django's cache framework](https://docs.djangoproject.com/en/3.2/topics/cache/).

Archmonger:
I agree that it should be async. Generally everything that supports Django 3.0+ only should be async, unless you're DB reliant (Django team is still working on async DB ORM).

[...]

PR #17, opened 2021-09-21
Advanced Websocket Features

Archmonger:
### New Features
- [x] Automatic WS reconnection (`settings.py:IDOM_WS_RECONNECT_TIMEOUT`)
- [x] Propogate WS down to components
- [x] Support authenticated components (Login the user and save the session if auth middleware is detected)
- [x] Pin `"idom-client-react": "^0.33.3"`
- [x] Update tests to add `websocket` as a component parameter
- [x] Update tests to use `render` shortcut instead of `HttpResponse`
- [x] Change `ALLOWED_HOSTS = ["*"]` in tests to allow LAN access 
- [x] Rename `IdomAsyncWebSocketConsumer` -> `IdomAsyncWebsocketConsumer`
- [x] Readme clean up. fixes, and updates


Archmonger:
I decided it would be best to propagate the entire websocket down, to fall in line with Django view convention of propagating down the HTTP request.

A benefit of this is that the developer would now be able to utilize any WS functionality (`scope`, `close()`, etc) from within their component.

If we need to add any features, can be added as props to the `IdomAsyncWebSocketConsumer`

Archmonger:
if I don't complete this PR by Thursday, I likely won't be available to pick it back up until October 4th.

Archmonger:
Reconnecting WS has been developed.

Currently works on client side disconnect.

### There are two caveats...
1) Reconnect is not graceful.
   - All IDOM modules are effectively "reset" to the first rendered state upon reconnect.
   - All variable values are lost.
   - We may need to implement a database store for linking a WS session to component state to resolve this.
2) Reconnections after server restart is broken due to Django IDOM design.
   - We only add modules to `IDOM_REGISTERED_COMPONENTS` upon template tag render, so all components are completely unknown after server restart (`"Unknown IDOM view ID"`).
   - We will need to develop the [previously mentioned `IDOM_COLLECT_OFFLINE`](https://github.com/idom-team/django-idom/issues/4) feature to resolve this.

Archmonger:
If approved, the last limb of Conreq blocking factors is `react-router` support.

Unfortunately I'm too much of a ReactJS scrub to PR that into IDOM Core, so I'll need your support on that.

src/django_idom/websocket_consumer.py
```diff
@@ -15,6 +18,12 @@
 _logger = logging.getLogger(__name__)
 
 
+@dataclass
+class WebSocketConnection:
+    scope: dict
+    disconnect: Callable
```
rmorshea, line 24: Can you make this a bit more explicit? Is this supposed to be `Callable[[], None]`?

src/js/src/index.js
```diff
@@ -4,28 +4,34 @@ import { mountLayoutWithWebSocket } from "idom-client-react";
 let LOCATION = window.location;
 let WS_PROTOCOL = "";
 if (LOCATION.protocol == "https:") {
-  WS_PROTOCOL = "wss://";
+	WS_PROTOCOL = "wss://";
```
rmorshea, line 7: It looks like there's a lot of unnecessary changes here. Can you make the indent here two spaces rather than three?

rmorshea (changes requested):
Besides those few things this looks good to me.

Archmonger:
Don't think I'll be able to tackle those changes until the 4th. On vacation, I'm not in my home state right now.

src/django_idom/websocket_consumer.py
Archmonger: Disconnect is a function we have within Django IDOM's consumer. So it'll never be none.

rmorshea:
No worries.

src/django_idom/websocket_consumer.py
rmorshea: What `Callable[[], None]` indicates is that the intended usage is `ws_conn.disconnect()` where the return value of `disconnect()` is `None`.

Given that it comes directly from Django IDOM's consumer, the signature would be `Callable[[int], Awaitable[None]]` with the intended use `await ws_conn.disconnect(code)`. This is fine, but we should probably have a sync version of this function too called `disconnect_soon`. This would be a a simple wrapper around `disconnect` that uses `asyncio.ensure_future`.

The interface here should then be:

```python
@dataclass
class WebSocketConnection:
    scope: dict
    disconnect: Callable[[int], typing.Awaitable[None]]
    disconnect_soon: Callable[[int], asyncio.Future[None]]
```

The implementation for `disconnect_soon` ought to look something like:

```python
def disconnect_soon(code: int) -> asyncio.Future[None]:
    return asyncio.ensure_future(disconnect(code))
```

Archmonger:
- Replaced WebSocketConnection `disconnect` with `close`, since close ultimately results in `disconnect` being called. 
- Added `view_id` to WebSocketConnection
- Fixed index.js spacing

src/django_idom/websocket_consumer.py
Archmonger: I don't think a sync helper will be needed since we'll always prefer calling async within our consumer.

rmorshea: You won't be able to call the async function from within a component directly though since that's a sync function. You'd need to do so in an event handler of component effect hook.

Archmonger: The django channels recommended way of using async within a sync context is using their conversion helper. We could simply recommend using it to minimize function wrappers.

```python 
from channels.db import database_sync_to_async, sync_to_async
```

---

With that in mind, has there been considerations for async components within IDOM core? It's theoretically possible for someone to have a long polling component action. Would be best if that doesn't lock up IDOM and Django Channels.

For example within Conreq I allow most things to be either via inspect.

```python
from inspect import iscoroutinefunction

async def sign_up(request):
    view = app.config.sign_up_view
    if iscoroutinefunction(view):
        return await view(request)
    return view(request)
```

src/django_idom/websocket_consumer.py
```diff
@@ -15,14 +18,37 @@
 _logger = logging.getLogger(__name__)
 
 
-class IdomAsyncWebSocketConsumer(AsyncJsonWebsocketConsumer):
+@dataclass
+class WebsocketConnection:
+    scope: dict
+    close: Callable[[Optional[int]], Coroutine[Any, Any, None]]
```
rmorshea, line 24: You should be able to just replace `Coroutine[Any, Any, None]` with `Awaitable[None]`

Archmonger: Alright, completed.

src/django_idom/websocket_consumer.py
rmorshea: So in this case I don't think those helpers from Django apply here since what we'd want is one that converts an async function to a sync one which is what `asyncio.ensure_future` is meant to do.

As for async components, in the past they actually were async. However there are two reasons they aren't now:

1. **Rendering the layout is a serial task** - technically you could render sibling components in parallel, but in 99% of cases that has no benefit at the cost of making things more complicated for the author to understand.
2. **You can use async functions in the [`use_effect`](https://idom-docs.herokuapp.com/docs/life-cycle-hooks.html#async-effects) hook** - an action like triggering the websocket to disconnect is a perfect example of a "side effect" that belongs withing the `use_effect` hook.

Archmonger: Apologies for the late reply.

Django ASGI-REF has the inverse of the previously mentioned function. For example: [async_to_sync](https://docs.djangoproject.com/en/3.2/topics/async/#asgiref.sync.sync_to_async). I don't think the developer manually disconnecting the websocket is common enough scenario to justify overcomplicating our function helpers. Frankly the only scenario I can think of it being utilizing is force disconnecting the WS after failed user authentication.

---

The discussion about async within IDOM should probably be moved to another thread but I'll quickly post my thoughts.

I agree that in most cases there isn't much benefit to async components, and they add complexity as you mentioned. That's pretty much the same conclusion that the Django team reached with async views. However, I think long term supporting both is ideal in order to prevent blocking/polling tasks from stalling IDOM.

The solution the Django team took was to the developer to decide whether to use sync or async views. To my knowledge, sync views are run in threads, and async views are handled through an event loop.

Archmonger:
As per my previous comment on your review, let me know what additional changes are needed in this PR.

rmorshea:
Been really busy lately. Will try to get to this next week.

Archmonger:
I updated the PR description.

Let me know if you need me to split this PR into separate parts to be more easily digestible.

tests/test_app/templates/base.html
```diff
@@ -1,23 +1,22 @@
...
+<body>
+  <h1>IDOM Test Page</h1>
+  <div>{% idom_component "test_app.components.HelloWorld" class="hello-world" %}</div>
+  <div>{% idom_component "test_app.components.Button" class="button" %}</div>
+  <div>{% idom_component "test_app.components.ParametrizedComponent" class="parametarized-component" x=123 y=456 %}
+  </div>
+  <div>{% idom_component "test_app.components.SimpleBarChart" class="simple-bar-chart" %}</div>
+</body>
+
+</html>
```
rmorshea, line 22: missing end of file new line

src/django_idom/config.py
```diff
@@ -10,6 +10,7 @@
 IDOM_BASE_URL = getattr(settings, "IDOM_BASE_URL", "_idom/")
 IDOM_WEBSOCKET_URL = IDOM_BASE_URL + "websocket/"
 IDOM_WEB_MODULES_URL = IDOM_BASE_URL + "web_module/"
+IDOM_WS_RECONNECT_TIMEOUT = getattr(settings, "IDOM_WS_RECONNECT_TIMEOUT", 604800)
```
rmorshea, line 13: So I need to rename this parameter in the JS client. This actually defines the ceiling for the delay between reconnection attempts. The client doesn't actually have a way to stop reconnection attempts at the moment :/

Maybe `IDOM_WS_MAX_RECONNECT_DELAY` would better communicate this.

rmorshea (changes requested):
LGTM - just minor nits.

Once I implement the `use_context` hook I think we'll be able to get rid of this websocket parameter and just allow the user to access the socket from anywhere:

```python
@component
def MyComponent():
    socket = use_context(WebSocketContext)
    ...
```

With everything else that needs to happen that's probably a ways off though so this should be fine for the time being.

tests/test_app/templates/base.html
Archmonger: On the note of EOF new lines, I'm going to make a PR to both DJI and Core to add a vscode configuration to automate things like this for new devs passing by.

src/django_idom/config.py
Archmonger: I'll also PR IDOM core to get a true "timeout".

rmorshea: I'd actually like to modify how reconnections happen so its more configurable. It should accept a callback that receives the last delay, and the total elapsed time as its arguments and returns the next delay.

Archmonger:
Side note...

We should also consider adding `pre-commit.ci` to all idom-team repos. It's an extremely handy utility for automating common code cleanup during the PR stage.

Since both of us use `black` on our repos, you can just take my fairly extensive [pre-commit.ci configuration file](https://github.com/Archmonger/Conreq/blob/a11f5748e9a41646bcce35fb70c259c6cac61905/.pre-commit-config.yaml).

Archmonger:
All comments have been resolved.

rmorshea:
> I see that IDOM Core is using [pre-commit] currently, but it's using a fairly minimal config.

From my experience, pre-commit has been most helpful in resolving issues that can be fixed automatically (e.g. black, sorting imports) which is why I've kept it fairly minimal. Linters like `flake8` and `pylint` that just report errors can be annoying if you want to introduce a commit that represents intermediate progress. So, including pre-commit hooks that add EOF new lines and strip out extra white space would be great to have though.

A longer-term solution to this would be to make an `idom-team/python-package-template` repo that pre-bakes all the standards we want to enforce.

rmorshea (approved):

Archmonger:
> A longer-term solution to this would be to make an `idom-team/python-package-template` repo that pre-bakes all the standards we want to enforce.

I recommend using the [repo-file-sync](https://github.com/BetaHuhn/repo-file-sync-action) GitHub action to automate standardized configurations across `idom-team` repos.

That's what we use at [Jazzband](https://github.com/jazzband/) alongside our [template repository](https://github.com/jazzband/.github).

PR #24, opened 2021-10-29
Auto populate IDOM component registry

Archmonger:
When Django first loads, attempt to populate the IDOM component registry.

This design was inspired by [`django_compressor`](https://github.com/django-compressor/django-compressor). See https://github.com/idom-team/django-idom/discussions/20 for more details.


Archmonger:
The only caveat to this design is it seems to inconsistently break `django-silk`.
Currently digging into root cause on that.

Archmonger:
Converting to draft while I debug django-silk compatibility

Archmonger:
It was too much of a hassle to fix `django-silk` non-HTTP request renders (silk's source is messy), so I opted to rely on regex to parse component names.

Functionally this implementation should be equivalent.

src/django_idom/utils.py
```diff
@@ -0,0 +1,125 @@
...
+                    and any(fnmatch(name, "*%s" % glob) for glob in extensions)
+                )
+
+        return templates
+
+    def _get_components(self, templates: Set) -> Set:
+        """Obtains a set of all IDOM components by parsing HTML templates."""
+        component_regex = re.compile(
+            r"{% *idom_component ((\"[^\"']*\")|('[^\"']*')).*%}"
+        )
```
rmorshea, line 102: Can you pull this regex out into a global variable and write some basic unit tests for it?

src/django_idom/utils.py
```diff
@@ -0,0 +1,125 @@
...
+                    if not name.startswith(".")
+                    and any(fnmatch(name, "*%s" % glob) for glob in extensions)
+                )
+
+        return templates
+
+    def _get_components(self, templates: Set) -> Set:
+        """Obtains a set of all IDOM components by parsing HTML templates."""
+        component_regex = re.compile(
+            r"{% *idom_component ((\"[^\"']*\")|('[^\"']*')).*%}"
```
rmorshea, line 101: Could you change `.*` to `.*?`

rmorshea (changes requested):

Archmonger:
I resolved the two comments, and additionally I've created a `tests` folder since we'll likely add more in the future.

Django mandates this test folder naming scheme, but it does feel a bit awkward though since our test app is also called `tests`. Might be something worth changing in a future PR?

src/django_idom/utils.py
```diff
@@ -0,0 +1,124 @@
...
+from importlib import import_module
+from typing import Set
+
+from django.template import engines
+from django.utils.encoding import smart_str
+
+from django_idom.config import IDOM_REGISTERED_COMPONENTS
+
+
+COMPONENT_REGEX_PATTERN = r"{% *idom_component ((\"[^\"']*\")|('[^\"']*')).*?%}"
```
rmorshea, line 14: You should be able to just do `COMPONENT_REGEX_PATTERN = re.compile(...)` so you don't have to re-compile every time its used.

tests/test_app/tests/test_regex.py
```diff
@@ -0,0 +1,41 @@
+from django.test import TestCase
+
+from django_idom.utils import COMPONENT_REGEX_PATTERN
+
+
+class RegexTests(TestCase):
+    def test_component_regex(self):
+        self.assertRegex(r'{%idom_component "my.component"%}', COMPONENT_REGEX_PATTERN)
```
rmorshea, line 8: Maybe this could be done in a for-loop?

rmorshea (changes requested):
LGTM besides a few more minor nits

rmorshea:
Once we get this in I think we can do a release.

Archmonger:
Completed!

rmorshea (approved):

PR #27, opened 2021-11-17
Component Initial HTML

Archmonger:
The idea of this PR is to allow for customizable initial HTML. 

This is useful for...
- Showing a simple loader/message in situations where a websocket connection could not be formed (Message display can be delayed via CSS).
- Can be used to run simple scripts on the component div.

This is different than an IDOM Core `idom.component` `fallback=...` that we have discussed due to the fact that this implementation fallback has no WS reliance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ In context this will look a bit like the following...
 </html>
 ```
 
-Keyworded parameters that can also be passed in include `class="example"` and `loader="example.html"`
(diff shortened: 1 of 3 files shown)


Archmonger:
I want to state that this PR is currently nonfunctional due to `mountViewToElement` wiping the inner contents of the component on WS connect.

Mostly opening this PR as a discussion as whether or not this concept is viable.

rmorshea:
I think this can be addressed once #581 is completed.

PR #47, opened 2022-01-12
Rename cache backend to "idom"

### LINKED ISSUE #46: Rename cache backend to "idom"
There's a high chance we will use our cache backend for more than just web modules in the future. It's best to rename the backend now while there still aren't many django-idom users yet.

Archmonger:
fix #46


src/django_idom/config.py
```diff
@@ -15,7 +15,7 @@
 _CACHES = getattr(settings, "CACHES", {})
 if _CACHES:
     if "idom_web_modules" in getattr(settings, "CACHES", {}):
-        IDOM_WEB_MODULE_CACHE = "idom_web_modules"
+        IDOM_WEB_MODULE_CACHE = "idom"
```
rmorshea, line 18: Should we rename `IDOM_WEB_MODULE_CACHE` to `IDOM_CACHE` as well?

Archmonger: Looks like I already handled this in PR #50 so I'm going to close this PR.

PR #49, opened 2022-01-12
Set IDOM_DEBUG_MODE to settings.py:DEBUG

### LINKED ISSUE #29: Set IDOM_DEBUG_MODE to settings.py:DEBUG
### Old Behavior

`IDOM_DEBUG_MODE` needed to be manually set

### New Behavior

Automatically set it based on `settings.py:DEBUG`

### Implementation Details

`IDOM_DEBUG_MODE = django.conf.settings.DEBUG`

### Code of Conduct

- [X] I agree to follow the [Code of Conduct](https://github.com/idom-team/django-idom/blob/main/CODE_OF_CONDUCT.md).

Archmonger:
fix #29

### DIFF
--- a/src/django_idom/config.py
+++ b/src/django_idom/config.py
@@ -1,9 +1,11 @@
+import os
 from typing import Dict
 
 from django.conf import settings
 from django.core.cache import DEFAULT_CACHE_ALIAS
 from idom.core.proto import ComponentConstructor
 
+os.environ["IDOM_DEBUG_MODE"] = str(int(settings.DEBUG))
 
 IDOM_REGISTERED_COMPONENTS: Dict[str, ComponentConstructor] = {}
 
--- a/src/django_idom/config.py
+++ b/src/django_idom/config.py
@@ -5,6 +5,7 @@ from django.conf import settings
 from django.core.cache import DEFAULT_CACHE_ALIAS
 from idom.core.proto import ComponentConstructor
 
+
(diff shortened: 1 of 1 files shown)


rmorshea:
The IDOM_DEBUG_MODE option may need some reworking. The value gets loaded on the first import of `idom` and can't be changed after that. At the moment many usages of `IDOM_DEBUG_MODE` in `idom` look like this:

```python
def some_function():
    ...
    
if IDOM_DEBUG_MODE.current:
    _some_function = some_function

    def some_function():
        # do some debug check
        _some_function()
```

Instead of:

```python
def some_function():
    if IDOM_DEBUG_MODE.current:
        # do some debug check
    ...
```

In retrospect trying to avoid `if IDOM_DEBUG_MODE.current` checks at runtime was a rather silly optimization.

Archmonger:
That's okay for now.
I'll open a PR in IDOM core to change this behavior, and then bump the IDOM version in this repo at a later date.

Archmonger:
From a Django IDOM perspective, I think this one can be merged.

Changing the behavior in core can be done separately.

rmorshea:
Setting `os.environ` at this point, won't make a different. `idom` initializes the debug setting once it's imported, so you'd need to set it before the first time `idom` is ever imported.

Archmonger:
Blocked on https://github.com/idom-team/idom/issues/582 so I am converting this to draft.

PR #50, opened 2022-01-12
Improve Caching Schema

### LINKED ISSUE #46: Rename cache backend to "idom"

Archmonger:
- [Use Django async caching API](https://docs.djangoproject.com/en/4.0/topics/cache/#asynchronous-support)
   - Note: This limits us to Django 4.0+, Python 3.8+.
- Remove async_lru_cache.
   - [Django's default cache engine is functionally equivalent](https://docs.djangoproject.com/en/4.0/topics/cache/#local-memory-caching).
- Fix #46 
- Use [`aiofile`](https://github.com/mosquito/aiofile) for async file operations.
- Protect against malicious file paths (ex `../../../etc/passwd`)
- [Use cache versioning to invalidate old web module files stored in cache](https://docs.djangoproject.com/en/4.0/topics/cache/#cache-versioning)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8, 3.9, 3.10]
+        python-version: ["3.8", "3.9", "3.10"]
     steps:
       - uses: actions/checkout@v2
       - uses: nanasess/setup-chromedriver@master
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.8, 3.9, 3.10]
     steps:
       - uses: actions/checkout@v2
       - uses: nanasess/setup-chromedriver@master
--- a/src/django_idom/config.py
+++ b/src/django_idom/config.py
@@ -1,7 +1,7 @@
 from typing import Dict
 
 from django.conf import settings
-from django.core.cache import DEFAULT_CACHE_ALIAS
+from django.core.cache import DEFAULT_CACHE_ALIAS, caches
 from idom.core.proto import ComponentConstructor
 
 
@@ -13,6 +13,6 @@ IDOM_WEB_MODULES_URL = IDOM_BASE_URL + "web_module/"
 IDOM_WS_MAX_RECONNECT_DELAY = getattr(settings, "IDOM_WS_MAX_RECONNECT_DELAY", 604800)
 
 if "idom" in getattr(settings, "CACHES", {}):
-    IDOM_CACHE = "idom"
+    IDOM_CACHE = caches["idom"]
 else:
-    IDOM_CACHE = DEFAULT_CACHE_ALIAS
+    IDOM_CACHE = caches[DEFAULT_CACHE_ALIAS]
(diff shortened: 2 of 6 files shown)


Archmonger:
This now uses `aiofile` as well, which means fetching IDOM web modules is fully async.

Should be performant enough to where we can put off #4 for a while.

Archmonger:
I had a concept to [use cache versioning to help invalidate old web module files stored in cache](https://docs.djangoproject.com/en/4.0/topics/cache/#cache-versioning), but unfortunately this doesn't work since the temp directory names are randomized, thus we have no replicable cache key.

rmorshea (approved):

src/django_idom/config.py
```diff
@@ -12,17 +12,7 @@
...
 IDOM_WS_MAX_RECONNECT_DELAY = getattr(settings, "IDOM_WS_MAX_RECONNECT_DELAY", 604800)
 
-_CACHES = getattr(settings, "CACHES", {})
-if _CACHES:
-    if "idom_web_modules" in getattr(settings, "CACHES", {}):
-        IDOM_WEB_MODULE_CACHE = "idom_web_modules"
-    else:
-        IDOM_WEB_MODULE_CACHE = DEFAULT_CACHE_ALIAS
+if "idom" in getattr(settings, "CACHES", {}):
+    IDOM_CACHE = "idom"
```
rmorshea, line 16: Maybe make `IDOM_CACHE = caches[IDOM_CACHE_NAME]` where `IDOM_CACHE_NAME = "idom"` so you don't need to key into `caches` everywhere you want to use it?

rmorshea:
> temp directory names are randomized

By default `IDOM_WEB_MODULES_DIR` is a temp dir, but that can be changed. For example you could set it using [appdirs](https://github.com/ActiveState/appdirs). That way it would be static.

src/django_idom/config.py
Archmonger: Great idea! I'll implement this.

rmorshea:
A note on changing `IDOM_WEB_MODULES_DIR`: since there may already be files there at the time you change the directory you should copy over any existing files into the new location.

Archmonger:
Is the default behavior within Core to not override existing files, if they have different contents?

rmorshea:
The default behavior does not overwrite. The code there is a little messy, but I just put up a PR to clean that up: https://github.com/idom-team/idom/pull/586

Archmonger:
Alright. I'll handle app directory stuff when idom-team/idom#586 is closed and made into a PyPI release.

In that case this PR is ready for merge. You need to make Python 3.7 tests non-required on your end for that to happen though.

rmorshea:
IDOM core can continue supporting 3.7, you'll just need to constrain the supported versions in `setup.py` for this package.

.github/workflows/test.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: ["3.8", "3.9", "3.10"]
```
rmorshea, line 18: This will require changing the supported version of Python in `setup.py`

rmorshea (changes requested):

.github/workflows/test.yml
Archmonger: Fixed.

rmorshea:
@Archmonger the latest version of IDOM has been released with the changes from https://github.com/idom-team/idom/pull/586. Is there anything else to do here?

Archmonger:
This PR should be feature complete, can't be merged without making 3.7 not required. I'm assuming that's in the project settings?

I'll make a separate PR to bump IDOM in here to the new version.

rmorshea (approved):

rmorshea:
Feel free to merge

PR #57, opened 2022-01-28
Bump IDOM to v0.35.4

Archmonger:
This resolves all major issues I had detected within Django IDOM, which means that we should be ready for a release.


rmorshea:
will take look shortly

rmorshea (approved):

Archmonger:
@rmorshea Scratch what I said. 0.35.4 is _very_ broken right now. 

Any state update I do on a child element results in `AssertionError: state mismatch during component update - key '61e47127-9d3f-4954-8235-4d99122134c1'!=4 `

PR #67, opened 2022-02-20
Fix 0.0.3 Publishing

Archmonger:
We forgot to bump `django-idom-client`.

We should think of some way of automating this, even if it's just a simple python script that manually edits all these values through regex.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ Types of changes are to be listed in this order
 
 ## [Unreleased]
 
+- Nothing (yet)
+
+## [0.0.3] - 2022-02-19
+
 ### Changed
 
 - Bumped the minimum IDOM client version to 0.36.3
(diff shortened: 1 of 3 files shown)


rmorshea (commented):
LGTM

Archmonger:
@rmorshea What's the recommended way of resolving this? The CI has automatically pushed to PyPI but the NPM build failed due to the missing tag.

I would assume this needs manual intervention.

rmorshea:
Yeah. I'll do a manual release.

rmorshea:
https://www.npmjs.com/package/django-idom-client/v/0.0.3

Archmonger:
Since the Django IDOM client is stored locally within this repo, do we even need to have it listed on NPM?

rmorshea:
Probably not. If you make a PR to remove that part of the release process and remove the `name` and `version` indicators from the `package.json` I can delete the package on NPM.

PR #68, opened 2022-02-20
Remove "django-idom-client" from NPM

Archmonger:
There is currently no benefit to having an NPM listing for `django-idom-client`.

For distribution, the JavaScript is stored within the PyPI release.
For developing, anyone working on `django-idom` will be rebuilding the client via `pip install -e .` anyways.


Archmonger:
@rmorshea Can we handle this one so I can push out a 0.0.4 release?

rmorshea (approved):

PR #71, opened 2022-04-04
Django IDOM v0.0.5

Archmonger:
The rename of `proto` to `types` broke Django IDOM, so we need a new release to fix this.


Archmonger:
Might want to consider setting up a daily workflow to test for things like this.

Archmonger:
Ah, it was failing. I just don't get email notifications for it since I'm not technically not the account running the workflow.

Not sure if I can [turn on notifications](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/notifications-for-workflow-runs) for this... I'm already set to be notified on "All Activity".

rmorshea (approved):

PR #76, opened 2022-06-02
Minor docs improvements

Archmonger:


docs/installation/index.md
```diff
@@ -1,8 +1,6 @@
-<!--dj-proj-start-->
+???+ tip "Learn Django before reading these docs!"
```
rmorshea, line 1: This sounds a bit harsh (like "just get good" in gaming). Maybe something like:

```suggestion
???+ tip "We recommend having basic knowledge of [Django](https://docs.djangoproject.com) before using Django IDOM"
```

Archmonger: I've changed the verbiage to "Learning Django first is recommended!".

Let me know if this works.

rmorshea (approved):

PR #77, opened 2022-06-23
Deferred static files

Archmonger:
This PR adds a `static_css` component, which reads the CSS file's from disk, stores the contents in cache (for performance purposes), and then puts the file contents into a `html.style` tag.

Currently, IDOM has no real way to "wait for CSS to load". This will lead to visual jank when attempting to use `html.link` to defer load CSS loading until a certain component exists. For example, visual jank occurs when using

```python
html.link({"rel": "stylesheet", "href": static("conreq/buttons.css")})
```

This also mitigates the need for the following (within Django only):
- https://github.com/idom-team/idom/issues/725

## Changelog
- Deferred static file components
- Docs for deferred static files
- Bump minimum IDOM version
- Create PR templates
- Sync issue form template with `idom-team/idom`
- Pin selenium to 4.2 for compatibility
- Bump django-idom to v1.1.0
- Docstring for component template tag
- Minor readme wordsmithing


src/django_idom/components.py
```diff
@@ -0,0 +1,32 @@
...
+    last_modified_time = os.stat(abs_path).st_mtime
+    cache_key = f"django_idom:css_contents:{static_path}"
+    file_contents = IDOM_CACHE.get(cache_key, version=last_modified_time)
+    if file_contents is None:
+        with open(abs_path, encoding="utf-8") as static_file:
+            file_contents = static_file.read()
+        IDOM_CACHE.delete(cache_key)
+        IDOM_CACHE.set(
+            cache_key, file_contents, timeout=None, version=last_modified_time
+        )
```
rmorshea, line 29: If it weren't for the fact that IDOM is currently using `is` for string and float comparisons you could do this pretty cleanly with `use_memo`:

```python
from pathlib import Path

def static_css(static_path: str):
    abs_path = find(static_path)
    if not abs_path:
        raise FileNotFoundError(
            f"Could not find static file {static_path} within Django's static files."
        )

    return html.style(
        use_memo(
            lambda: Path(abs_path).read_text(encoding="utf-8"),
            dependencies=[abs_path, os.stat(abs_path).st_mtime],
        )
    )
```

I've already been thinking about special casing comparisons for ints, floats, strings etc to use `==` instead of `is` (which is basically how the `===` operator works in JS from what I can tell) and this example makes a good case for making that change.

rmorshea: https://github.com/idom-team/idom/issues/773

Archmonger: @rmorshea `use_memo` is not multiprocessing compatible, while `IDOM_CACHE` is.

rmorshea (approved):

Archmonger:
@rmorshea This PR is still a draft, I have a bit more to work on.

I will mark as "Ready for review" when complete.

src/django_idom/components.py
```diff
@@ -0,0 +1,41 @@
+import os
+
+from django.contrib.staticfiles.finders import find
+from idom import component, html
+
+from django_idom.config import IDOM_CACHE
+
+
+@component
+def static_css(static_path: str):
```
Archmonger, line 10: This brings up a bit of a debate.

I mentioned in the past that I prefer if we follow PEP8 styling for component names instead of React styling. This is mostly to avoid pissing off users that use Python style linters.

Whatever we decide on this will likely decide what we do in the future for any prefab components.

Archmonger:
@rmorshea Would you have time to review this weekend?

rmorshea:
I'll try to get to this today or tomorrow.

src/django_idom/components.py
Archmonger: Going to assume we're pushing forward with PascalCase so I'm going to mark this as resolved.

rmorshea:
apologies. i will make time tomorrow

src/django_idom/components.py
Archmonger: While trying to convert Conreq component names to PascalCase, the pylint warnings started to bother me so I re-opened this.

.github/ISSUE_TEMPLATE/issue-form.yml
```diff
@@ -13,11 +13,4 @@ body:
...
   validations:
-    required: false
-- type: textarea
-  attributes:
-    label: Work Items
-    description: |
-      An itemized list or detailed description of the work to be done to based on the proposed actions above.
-  validations:
-    required: false
+    required: false
```
rmorshea, line 16:
```suggestion
    required: false

```

src/django_idom/components.py
```diff
@@ -0,0 +1,51 @@
...
+    abs_path = find(static_path)
+    if not abs_path:
+        raise FileNotFoundError(
+            f"Could not find static file {static_path} within Django's static files."
+        )
+
+    # Fetch the file from cache, if available
+    last_modified_time = os.stat(abs_path).st_mtime
+    cache_key = f"django_idom:static_contents:{static_path}"
+    file_contents = IDOM_CACHE.get(cache_key, version=last_modified_time)
```
rmorshea, line 42: Could you add a comment about the fact that this is preferable to `use_memo` because the cache works across-processes?

tests/test_app/static/static-css-test.css
```diff
@@ -0,0 +1,3 @@
+#static-css button {
+    color: rgba(0, 0, 255, 1);
+}
```
rmorshea, line 3:
```suggestion
}

```

tests/test_app/static/static-js-test.js
```diff
@@ -0,0 +1,3 @@
+let el = document.body.querySelector("#static-js");
+el.textContent = "StaticJS: True";
+el.dataset.success = "true";
```
rmorshea, line 3:
```suggestion
el.dataset.success = "true";

```

rmorshea (changes requested):
Just a few minor things here and there. Besides that, LGTM!

src/django_idom/components.py
Archmonger: @rmorshea Let's discuss this before PR merge.

This is important to hash out now since this is our first production-capable prefab component.

rmorshea: I think for the sake of (1) alone, we should probably use snake case going forward.

rmorshea: I also realized that making them camel case makes it seem like component functions can be used in type annotations, which they cannot.

Archmonger:
Can't merge without re-approval.

rmorshea (approved):

PR #80, opened 2022-06-26
Allow Django ORM usage within components

### LINKED ISSUE #79: Django ORM does not work well with IDOM
### Current Situation

Currently, the Django ORM will shoot out a `SynchronousOnlyOperation` if used within a IDOM component. This is caused by Django's current limitation surrounding mixed sync-async contexts.

### Proposed Actions

Choose one of the following

1) Create a React-style hooks to perform ORM [CRUD](https://en.wikipedia.org/wiki/Create,_read,_update_and_delete)
2) Set `DJANGO_ALLOW_ASYNC_UNSAFE` when within a component rendering context.
3) Create a hook that utilizes `database_sync_to_async` that can perform ORM queries

Archmonger:
- #79


rmorshea:
While I understand the aversion towards loading things from the database in effects, this is actually very common and generally recommend in React. That's because you don't want to block rendering while you're reading data, and you want to keep side effects like writing data to the database, inside a `use_effect` hook.

Typically, the way people handle this in react is to create a hook like this

```python
data, loading, error = use_query(query)

if loading:
    return loading_spinner()
if error:
    return error_message(error)

return the_actual_view(data)
```

Where the implementation of `use_query` looks something like:

```python
def use_query(query):
    data, set_data = use_state(None)
    loading, set_loading = use_state(True)
    error, set_error = use_state(None)

    @use_effect
    async def load_data():
        set_loading(True)
        try:
            set_data(fetch_from_database(query)
        except Exception as error:
            set_error(error)
        finally:
            set_loading(False)

    return data, loading, error
```

An example of such a hook can be found in the popular GraphQL client [Apollo](https://www.apollographql.com/docs/react/get-started/#step-5-fetch-data-with-usequery).

My recommendation would be to create a `use_orm` hook that functions in a similar way rather than allowing people to make queries in the main body of their components instead of in effects.

Archmonger:
I believe the `use_orm` hook should live alongside this implementation. If the user intentionally doesn't mind the blocking behavior of ORM queries, I say let them have it.

For example, in Conreq's circumstance, there's a lot of situations where I'm absolutely reliant on ORM queries for each render. Having it built into the component vastly simplifies the design.

rmorshea:
I guess I understand that. However, given that loading and mutating database resources should generally be done in effects, we ought to give people an easier way to do that since it's tedious to write out the loading/error/success states every time you need to access the DB in a component.

Archmonger:
The Django ORM doesn't work well in effects either due to similar protection schemas.

Also, I'm about 80% certain we can't wrap the Django ORM in a `use_orm` hook due to those same protections. We very likely would either need heavy monkey patching on the ORM, or upstream PRs on `django`

rmorshea:
Ok, I don't think I understood the issue here. The fact that we're running in an event loop at all makes Django unhappy. Would something like this work to avoid the `SynchronousOnlyOperation` error?

```python
@component
def sample():

    @use_effect
    @async_to_sync  # <-- I'm thinking that this might make Django happy in the thread_sensitive=True mode
    async def do_orm_work():
        Model.objects.all()
        ...
```

Archmonger:
I haven't tested the decorator directly on the effect, but it certainly does work [when used inside the effect](https://github.com/Archmonger/Conreq/blob/e44aed1c1f522c4cdaeb40c56f4f66e7e5f21407/conreq/_core/app_store/components.py#L32)

rmorshea:
Also, is it the case that the `SynchronousOnlyOperation` only happens when a query set is executed? If so, then I think there might be a decent way to do this. I'm thinking the interface for queries (still have to think about create/update operations) might be:

```python
data, loading, error = use_query(Model.objects.all())
```

Inside `use_query` we would force the `QuerySet` to execute inside an effect wrapped in `async_to_sync`.

Archmonger:
I will need to double check at what point during execution the exception is raised. I will get back to you on that.

Archmonger:
And, if the exception is caused when `.all()` is called, we _might_ be able to get away with an interface like this:

```python
def use_query(query: Callable, *args, **kwargs):
   ...

data, loading, error = use_query(Model.objects.all, example_val, value=other_example)
```

rmorshea:
We'd take an approach similar to Apollo's [`useMutation` hook](https://www.apollographql.com/docs/react/data/mutations/#executing-a-mutation) to handle ORM object creation and updating.

If this is truly blocking we can push forward, but ultimately I'd like to remove this functionality in favor of hooks. We should brainstorm the interface for these hooks in a separate issue.

Archmonger:
That useMutation API frankly looks quite ugly. That would be a really tough bullet for Django developers to swallow.

Archmonger:
Is it worth pushing out this PR just to enable Django ORM behavior to be similar to the other frameworks IDOM supports?

If not, I'll close this PR.

rmorshea (approved):
I can see it taking a bit for us to hammer out the details of a hook-based alternative. We can do this as a short-term solution if you think that's necessary.

Archmonger:
If we're going to have to back this out at some point, I think it might be more realistic to just document how to use `use_effect` with `database_sync_to_async` as a temporary solution.

PR #84, opened 2022-07-02
use_sync_to_async Hook

Archmonger:
This hook is best way I can think of developing ORM query support, at least in a way that Django developers would understand.

Potentially closes the following
- #79
- #80

### DIFF
--- a/docs/features/hooks.md
+++ b/docs/features/hooks.md
@@ -1,8 +1,38 @@
-# Django Hooks
-
 ???+ tip "Looking for more hooks?"
 
     Check out the [IDOM Core docs](https://idom-docs.herokuapp.com/docs/reference/hooks-api.html?highlight=hooks) on hooks!
+
+## Use Sync to Async
+
+This is the suggested method of performing ORM queries when using Django IDOM.
+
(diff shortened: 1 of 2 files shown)


Archmonger:
@rmorshea Let me know what you think about this implementation.

docs/features/hooks.md
```diff
@@ -1,9 +1,41 @@
...
+    categories, set_categories = hooks.use_state(None)
+
+    @hooks.use_sync_to_async
+    def get_categories():
+        if categories:
+            return
+        set_categories(Category.objects.all())
+
+    if not categories:
+        return html.h2("Loading...")
```
rmorshea, line 28: Part of the reason for the `data, loading, error` pattern is that the `if not categories` check does not indicate whether the query is loading - it could be that there are no categories to show, that loading categories failed, or that categories are indeed loading. You also need a convenient way to re-execute queries that are related. For example, triggering an `add_categories` database operation should cause the `get_categories` effect to re-run. Whatever solution we provide should make it easier to distinguish between the possible states a query or mutation is in, as well as re-trigger dependent queries.

PR #86, opened 2022-07-02
use_query and use_mutation

rmorshea:
# Description

Adds `use_query` and `use_mutation` hooks. I haven't tested to see that these work as expected, but they get across the interface that I think we should provide for writing database operations.

# Checklist:

-   [ ] Tests have been included for all bug fixes or added functionality.
-   [ ] The `changelog.rst` has been updated with any significant changes, if necessary.
-   [ ] GitHub Issues which may be closed by this PR have been linked.


Archmonger:
- Recursively for loop iterate through each object value to force run the query while within the hook
- Have a flag to disable `immediate_fetch`
- Double check the re-run workflow to ensure we aren't unnecessarily performing ORM queries

rmorshea:
@Archmonger so the implementation I went with uses `Model.get_deferred_fields` and just called `getattr` on all the field names it returns. I'm pretty sure this does what we want. I also require that if `fetch_deferred_fields=True`, then the query function must return a `QuerySet` or `Model` because it would be too hard to try and discover `Model` instances in arbitrary data structures in order to fetch their deferred fields.

rmorshea:
@Archmonger I've got to switch and work on some other things for a bit. If you're able to help get this over the finish by adding some docs or tests that would be great.

Archmonger:
Will do, but I am currently out on vacation. I'll get to it sometime after the 20th of this month.

src/django_idom/hooks.py
```diff
@@ -30,3 +40,145 @@ def use_websocket() -> IdomWebsocket:
...
+    given_query = query
+    query, _ = use_state(given_query)
+    if given_query is not query:
+        raise ValueError(f"Query function changed from {query} to {given_query}.")
+
+    data, set_data = use_state(UNDEFINED)
+    loading, set_loading = use_state(True)
+    error, set_error = use_state(None)
+
+    @use_callback
```
Archmonger, line 69: Is there a functional reason for having this be `use_callback`?

src/django_idom/hooks.py
```diff
@@ -30,3 +40,145 @@ def use_websocket() -> IdomWebsocket:
...
         raise RuntimeError("No websocket. Are you running with a Django server?")
     return websocket
+
+
+_REFETCH_CALLBACKS: DefaultDict[FunctionType, set[Callable[[], None]]] = DefaultDict(
+    set
+)
+
+
+_Data = TypeVar("_Data")
```
Archmonger, line 50: Functionally this seems redundant to me, is there a purpose behind this typevar?

rmorshea: Yes, this is used to parametrize `Query`. This allows us to say that the return value of the `Query.data` will be the same type as is returned by the `query` function.

src/django_idom/hooks.py
rmorshea: In the context of the hook no, but if we don't the identity of the function will change on every render. Then, If this callback is a dependency in the user's code (e.g. an effect) we don't want to cause unnecessary renders.

This is common in hook libraries. For example, the `set_state` callback from the `use_state` hook does not change identity after a render. In application code, `use_callback` shows up less commonly and is just a way to improve performance where necessary since each usage has a memory cost.

Archmonger:
While I was writing out the docs I had a thought. Is there really any point to having separate `use_query` and `use_mutation` hooks?

I feel like the `use_query` hook could easily fill the role of both. For example, if the user doesn't return anything within their query function, we can just assume it's a mutation.

docs/features/hooks.md
```diff
@@ -2,6 +2,117 @@
...
+    from example_project.my_app.models import TodoItem
+    from idom import component, html
+    from django_idom.hooks import use_mutation
+
+    def add_item(text: str):
+        TodoItem(text=text).save()
+
+    @component
+    def todo_list():
+        item_mutation = use_mutation(add_item)
```
rmorshea, line 70: A better example here would be to display the todo items in this component too. You could then demonstrate `use_mutation(add_item, refetch=get_items)`.

The purpose of refetch is to re-run any queries that might have been impacted by a mutation. For example, in this case, after running `add_item` to the database, your prior `get_items` query would be stale and should be re-executed to get the latest items. This is especially useful if `get_items` was used in a different component since, `use_mutation(add_items, refetch=get_items)` will automatically tell components that used `use_query(get_items)` to re-render regardless of where they are in your layout.

src/django_idom/hooks.py
```diff
@@ -30,3 +39,145 @@ def use_websocket() -> IdomWebsocket:
...
+        def thread_target():
+            try:
+                mutate(*args, **kwargs)
+            except Exception as e:
+                set_loading(False)
+                set_error(e)
+            else:
+                set_loading(False)
+                set_error(None)
+                for query in (refetch,) if isinstance(refetch, Query) else refetch:
```
rmorshea, line 151: I incorrectly implemented this check. It should actually be `isinstance(refetch, FunctionType)`

rmorshea, line 151:
```suggestion
                for query in (refetch,) if callable(refetch) else refetch:
```

rmorshea:
Let me know if you want to chat about `use_mutation` and the purpose of refetch.

docs/features/hooks.md
Archmonger: I planned on having that more expansive example within a question collapse. I think it's too much info to give at once if that's at a top level.

rmorshea: Got it. That makes sense.

docs/features/hooks.md
```diff
@@ -2,6 +2,117 @@
...
+
+        if item_mutation.loading:
+            mutation_status = html.h2("Adding...")
+        elif item_mutation.error:
+            mutation_status = html.h2("Error when adding!")
+        else:
+            mutation_status = ""
+
+        def submit_event(event):
+            # TODO: How do I get the text from this input?
```
Archmonger, line 80: By the way, what's the appropriate way of getting an input's text in IDOM?

rmorshea: I can't remember why, but you need to add a `value` to the input and handle `onChange`:

```python
value, set_value = use_state("")
return html.input({"onChange": lambda event: event["target"]["value"], "value": value})
```

Archmonger:
Note; Need to verify whether we are properly [closing database connections](https://andrewkowalik.com/posts/django-database-connnections-in-kafka/) in our new hooks.

We don't want these connections to potentially be re-used by other threads, otherwise data loss can occur. A quick n dirty solution could be using [`database_sync_to_async`](https://channels.readthedocs.io/en/stable/topics/databases.html#database-sync-to-async).

src/django_idom/hooks.py
```diff
@@ -30,3 +39,145 @@ def use_websocket() -> IdomWebsocket:
...
+
+
+_Data = TypeVar("_Data")
+_Params = ParamSpec("_Params")
+
+
+def use_query(
+    query: Callable[_Params, _Data],
+    *args: _Params.args,
+    fetch_deferred_fields: bool = True,
```
Archmonger, line 56: This arg causes the following Pylance warning:

Keyword parameter "fetch_deferred_fields" cannot appear in signature after ParamSpec args parameter
Pylance [reportGeneralTypeIssues](https://github.com/microsoft/pylance-release/blob/main/DIAGNOSTIC_SEVERITY_RULES.md#diagnostic-severity-rules)

rmorshea: Dang. Well, the only way to work around that is to add a second function which implements the `fetch_deferred_fields=False` behavior. I can't really think of a good name for it though.

Archmonger: I don't think this option should even exist. Not fetching deferred fields can completely break the ORM across all of Django, so I don't see any value in the toggle.

rmorshea:
On rereading the code I wrote I found a bunch of bug and removed `fetch_deferred_fields` along with the `UNDEFINED` sentinel.

src/django_idom/hooks.py
```diff
@@ -30,3 +54,117 @@ def use_websocket() -> IdomWebsocket:
...
+            else:
+                set_loading(False)
+                set_error(None)
+                for query in (refetch,) if callable(refetch) else refetch:
+                    for callback in _REFETCH_CALLBACKS.get(query) or ():
+                        callback()
+
+        # We need to run this in a thread so we don't prevent rendering when loading.
+        # We also can't do this async since Django's ORM doesn't support this yet.
+        Thread(target=thread_target, daemon=True).start()
```
Archmonger, line 140: I believe we may as well do this async, using [`database_sync_to_async`](https://channels.readthedocs.io/en/stable/topics/databases.html).

This would also cover us in automatically closing out the connection to prevent other threads from reusing it.

src/django_idom/hooks.py
```diff
@@ -41,84 +50,72 @@ def use_websocket() -> IdomWebsocket:
...
-                set_error(e)  # type: ignore
+                set_error(e)
                 return
+            finally:
+                set_should_execute(False)
 
-            if isinstance(query_result, QuerySet):  # type: ignore
-                if fetch_deferred_fields:
-                    for model in query_result:
-                        _fetch_deferred_fields(model)
```
Archmonger: Is there a reason why this was removed?

rmorshea: my bad, refactored anticipating a section function and forgot to put this back. Will fix that tomorrow.

rmorshea:
So I added a basic test. I'm not really sure why, but sometimes it fails because database is locked.

Archmonger:
I increase the database query timeout (it's fairly low by default), which resolved the `database is locked` error.

The current issue with the tests is that sometimes the item does not successfully update after a `use_mutation` execution.

rmorshea:
So it turns out there's a JS error that happens when applying a JSON patch generated by the server. I think this is, yet again, a bug with the Python JSON patch library. I'm growing increasingly frustrated with it and there's been no response to [this issue](https://github.com/stefankoegl/python-json-patch/issues/138) I raised over 6 months ago. I see a couple options to move forward:

1. Try and figure out the underlying issue and fix it
2. Find a more reliable JSON patch implementation. There are no other native Python implementations so we'd need a C or Rust extension.
3. Use a different solution altogether.

Archmonger:
Can we just perform the patch on the JS side? Last I checked there are some well maintained libs in that end.

rmorshea:
1. The main problem, if this is related to the prior issue, is that I don't really know how to it. The codebase is pretty hard to follow.

2. We're already doing this. The server generates a patch and the client applies that patch to its copy of the model.

3. I'd be open to doing some benchmarking and seeing if we could just get rid if it, but I suspect that if you have to re-render a large page from the root, that computing and sending the patch, vs sending the whole update, is worth it.

rmorshea:
BTW we're [already using](https://github.com/idom-team/idom/blob/main/src/idom/core/_fixed_jsonpatch.py) my proposed fix from the issue.

rmorshea:
I'll have to check, but perhaps switching to an older version of `jsonpatch` might fix things in the short-term.

Archmonger:
Although the tests are working, this still seems to be broken (tested via manually typing/clicking)

rmorshea:
Tracking the JSON Patch issue [here](https://github.com/idom-team/idom/issues/801). Also have a [PR](https://github.com/idom-team/idom/pull/802) up to resolve it. Just need to update tests.

Also, discovered [this](https://github.com/idom-team/idom/issues/799) while working on the tests. For some reason `onChange` is the only event that gets the accurate `event.target.value` if the input's value is "controlled".

rmorshea:
Ironically, in trying to remove JSON patch, I think I found the problem. I was [mutating something](https://github.com/idom-team/idom/blob/f593ee3067cac004ca07e4c8378f87f8c45341f3/src/idom/core/layout.py#L234-L236) I shouldn't have been which was screwing up the diff.

Archmonger:
I'm not sure what the new API is for `create_context` and `use_context` now.

rmorshea:
So I thought that change had fixed the JSON patch problem, but I don't think it has actually.

Archmonger:
IDOM 4.0 release might be even more broken than previous. Random VDOM elements simply don't render what so ever, even under normal circumstances.

Archmonger:
Looks like the VDOM for the previously mentioned broken test case is valid? If that's the case then this issue is almost certainly layout module related.

```python
{'attributes': {'id': 'django-css'},
 'children': [django_css(2131d8d8b80, static_path='django-css-test.css'),
              {'attributes': {'style': {'display': 'inline'}},
               'children': ['django_css: '],
               'tagName': 'div'},
              {'children': ['This text should be blue.'], 'tagName': 'button'},
              {'tagName': 'hr'}],
 'tagName': 'div'}
```

Archmonger:
Seems like JSON patch is still broken

```
client.js:30213 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'onChange')
    at applyOperation (client.js:30213:22)
    at Object.applyPatch (client.js:30250:22)
    at applyNonMutativePatch (client.js:30649:20)
    at client.js:30633:20
    at client.js:31298:48
````

---

On a different note, I am going to merge the playwright testing suite into my `view_to_component` PR.

**Comment by rmorshea** [comment]
@Archmonger it looks like the doc tests are failing. Getting 503 when checking links on the Django docs. Maybe getting rate limited since the links seem to be valid?

Also, I'm unable to reproduce the error you're seeing under the [these changes](https://github.com/idom-team/idom/pull/809) or under the latest 0.40.1 release of IDOM.

**Comment by Archmonger** [comment]
Error occurs when toggling any of the checkboxes on the use_mutation list elements.

**Comment by rmorshea** [comment]
The docs error looks real actually. Django's website is just down for some reason:

image

**Comment by Archmonger** [comment]
Yep, just tested that as well. Looks like the server they use to host dev docs is down. I might switch the docs off of the dev branch if this becomes a reoccurring issue.

**Comment by rmorshea** [comment]
Ok, am able to repro the problem on 0.40.1 albeit with a slightly different error message. Under proposed changes in [bypass-jsonpatch](https://github.com/idom-team/idom/pull/809) that problem goes away.

**Comment by Archmonger** [comment]
I just tested out the current branch again, with `bypass-jsonpatch`. The following errors were given on three separate mouse clicks.

```js
client.js:29997 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'id')
    at Object.replace (client.js:29997:26)
    at applyOperation (client.js:30206:60)
    at Object.applyPatch (client.js:30250:22)
    at applyNonMutativePatch (client.js:30649:20)
    at client.js:30633:20
    at client.js:31298:48
replace @ client.js:29997
applyOperation @ client.js:30206
applyPatch @ client.js:30250
applyNonMutativePatch @ client.js:30649
(anonymous) @ client.js:30633
(anonymous) @ client.js:31298
Promise.then (async)
socket.onmessage @ client.js:31298

client.js:30213 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'children')
    at applyOperation (client.js:30213:22)
    at Object.applyPatch (client.js:30250:22)
    at applyNonMutativePatch (client.js:30649:20)
    at client.js:30633:20
    at client.js:31298:48
applyOperation @ client.js:30213
applyPatch @ client.js:30250
applyNonMutativePatch @ client.js:30649
(anonymous) @ client.js:30633
(anonymous) @ client.js:31298
Promise.then (async)
socket.onmessage @ client.js:31298

client.js:3636 Warning: Encountered two children with the same key, `9`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
    at ul
    at StandardElement (http://127.0.0.1:8000/static/django_idom/client.js:31094:28)
    at Element (http://127.0.0.1:8000/static/django_idom/client.js:31076:20)
    at StandardElement (http://127.0.0.1:8000/static/django_idom/client.js:31094:28)
    at Element (http://127.0.0.1:8000/static/django_idom/client.js:31076:20)
    at div
    at StandardElement (http://127.0.0.1:8000/static/django_idom/client.js:31094:28)
    at Element (http://127.0.0.1:8000/static/django_idom/client.js:31076:20)
    at StandardElement (http://127.0.0.1:8000/static/django_idom/client.js:31094:28)
    at Element (http://127.0.0.1:8000/static/django_idom/client.js:31076:20)
    at StandardElement (http://127.0.0.1:8000/static/django_idom/client.js:31094:28)
    at Element (http://127.0.0.1:8000/static/django_idom/client.js:31076:20)
    at StandardElement (http://127.0.0.1:8000/static/django_idom/client.js:31094:28)
    at Element (http://127.0.0.1:8000/static/django_idom/client.js:31076:20)
    at Layout (http://127.0.0.1:8000/static/django_idom/client.js:31060:19)
```

Archmonger:
I did some inspection of the code in my venv, seems like the `git+` install wasn't overriding my local `idom`. I manually applied the patch to `serve.py` and it resolved the issue.

src/django_idom/websocket/consumer.py
```diff
@@ -23,6 +22,9 @@ class IdomAsyncWebsocketConsumer(AsyncJsonWebsocketConsumer):
     """Communicates with the browser to perform actions on-demand."""
 
     async def connect(self) -> None:
+        # this triggers AppRegistryNotReady exception in manage.py if at root level
+        from channels.auth import login
+
```
Archmonger, line 27: I tested on my branch that this is not required. Should revert this.

src/js/package.json
```diff
@@ -16,7 +16,7 @@
     "rollup-plugin-replace": "^2.2.0"
   },
   "dependencies": {
-    "idom-client-react": "^0.37.2",
+    "idom-client-react": "^0.40.1",
```
Archmonger, line 19: Need to bump client version

tests/test_app/tests/test_components.py
```diff
@@ -1,98 +1,104 @@
...
-
-        def test_authorized_user(self):
+from django.test import TransactionTestCase
+from playwright.sync_api import TimeoutError, sync_playwright
+
+
+class TestIdomCapabilities(
+    ChannelsLiveServerTestCase,
+    # using the normal TestCase caused the database to lock up after tests
+    TransactionTestCase,
```
Archmonger, line 13: `TransactionTestCase` is not needed here. `ChannelsLiveServerTestCase` inherits from `TransactionTestCase `

Archmonger (changes requested):

tests/test_app/components.py
```diff
@@ -1,22 +1,25 @@
...
+    return html.h1({"id": "hello-world"}, "Hello World!")
 
 
 @idom.component
 def button():
     count, set_count = idom.hooks.use_state(0)
-    return idom.html.div(
-        idom.html.button(
+    return html.div(
+        html.button(
```
Archmonger, line 18: Let's leave `idom.html` -> `html` changes for the `view_to_component` branch, just to reduce merging complexity.

src/django_idom/hooks.py
```diff
@@ -30,3 +58,124 @@ def use_websocket() -> IdomWebsocket:
...
+    refetch: Callable[[], None]
+
+
+@dataclass
+class Mutation(Generic[_Params]):
+    execute: Callable[_Params, None]
+    loading: bool
+    error: Exception | None
+    reset: Callable[[], None]
+
```
Archmonger, line 164: Should move this to `types.py`

src/django_idom/hooks.py
```diff
@@ -17,7 +17,7 @@
     Callable[..., Callable[..., Awaitable[Any]]],
     _database_sync_to_async,
 )
-WebsocketContext: Context[IdomWebsocket | None] = create_context(None)
```
rmorshea: You can use the `|` operator if you add `from __future__ import annotations` to the top of the file.

Archmonger:
Should I commit the `from channels.auth import login` and `TransactionTestCase` changes I suggested above?

rmorshea:
Go for it!

Archmonger (approved):

Archmonger:
LGTM, ready for merge.

PR #96, opened 2022-09-16
Allow whitespace characters within component regex

Archmonger:
## Description

I realized some handlebars formatters will add indentation and newlines. Therefore, I've made the component regex a bit more robust to handle those situations.

## Checklist:

-   [X] Tests have been included for all bug fixes or added functionality.
-   [X] GitHub Issues which may be closed by this PR have been linked.

### DIFF
--- a/src/django_idom/utils.py
+++ b/src/django_idom/utils.py
@@ -13,7 +13,9 @@ from django.utils.encoding import smart_str
 from django_idom.config import IDOM_REGISTERED_COMPONENTS
 
 
-COMPONENT_REGEX = re.compile(r"{% *component +((\"[^\"']*\")|('[^\"']*'))(.*?)%}")
+COMPONENT_REGEX = re.compile(
+    r"{%\s*component\s*((\"[^\"']*\")|('[^\"']*'))\s*((.*?|\s*?)*)\s*%}"
+)
 _logger = logging.getLogger(__name__)
 
 
--- a/src/django_idom/utils.py
+++ b/src/django_idom/utils.py
@@ -13,6 +13,7 @@ from django.utils.encoding import smart_str
 from django_idom.config import IDOM_REGISTERED_COMPONENTS
 
 
+_logger = logging.getLogger(__name__)
 _component_tag = r"component"
 _component_path = r"((\"[^\"']*\")|('[^\"']*'))"
 _component_kwargs = r"((.*?|\s*?)*)"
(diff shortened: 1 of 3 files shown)


src/django_idom/utils.py
```diff
@@ -13,8 +13,19 @@
 from django_idom.config import IDOM_REGISTERED_COMPONENTS
 
 
-COMPONENT_REGEX = re.compile(r"{% *component +((\"[^\"']*\")|('[^\"']*'))(.*?)%}")
 _logger = logging.getLogger(__name__)
+_component_tag = r"component"
+_component_path = r"((\"[^\"']*\")|('[^\"']*'))"
+_component_kwargs = r"((.*?|\s*?)*)"
```
rmorshea, line 19: While we're at it. Can we add names to these groups with `(?P<...>)`? Will just help out readability when matching.

rmorshea:
In usages of the regex pattern you can use `pattern.match(...).group(group_name)` instead of grabbing them by index.

Archmonger:
`pattern.match(...).group(group_name)` only works with single objects. To "find all" I needed to use `finditer` and `groupdict`.

I don't believe I'm using the index anywhere anymore.

src/django_idom/utils.py
```diff
@@ -102,11 +113,14 @@ def _get_components(self, templates: set[str]) -> set[str]:
         for template in templates:
             with contextlib.suppress(Exception):
                 with open(template, "r", encoding="utf-8") as template_file:
-                    match = COMPONENT_REGEX.findall(template_file.read())
-                    if not match:
+                    regex_iterable = COMPONENT_REGEX.finditer(template_file.read())
+                    if not regex_iterable:
                         continue
```
rmorshea, line 118: This is always going to be true. If the iterable is empty `update()` will do nothing anyway.
```suggestion
```

rmorshea, line 118: No need for this check. The `update()` will do nothing if the iterable is empty
```suggestion
```

rmorshea (changes requested):

src/django_idom/utils.py
```diff
@@ -102,11 +113,14 @@ def _get_components(self, templates: set[str]) -> set[str]:
...
                     components.update(
-                        [group[0].replace('"', "").replace("'", "") for group in match]
+                        [
+                            match.groupdict()["path"].replace('"', "").replace("'", "")
```
rmorshea, line 121: I'm pretty sure you can just do this if you only need one named group:
```suggestion
                            match.group("path").replace('"', "").replace("'", "")
```

rmorshea (approved):

PR #98, opened 2022-09-22
Rework view_to_component

Archmonger:
## Description _(NEW)_

This PR reworks _view_to_component_ to prevent needing to constantly pass in frequently used parameters.

## Description _(OLD)_

In Conreq, I am relying on several `view_to_component` URLs to exist prior to the first render. That meaning, the `compatibility` registration needs to occur prior to the component's first time rendering.

By moving the iframe registration outside of the hook, this change allows me to arbitrarily call `view_to_component(my_view, compatibility=True)` as a "preloading" step during Conreq initialization to ensure the URL exists beforehand.

Additionally, a nice benefit of this is now type hints will work. They were previously obfuscated due to the `@component` decorator.

In my opinion, this feels less obtrusive than a separate API just to register iframes.

## Checklist:

-   [X] The `changelog.rst` has been updated with any significant changes, if necessary.
-   [X] GitHub Issues which may be closed by this PR have been linked.


Archmonger:
Current implementation does not work as expected. Might need to think about this some more.

src/django_idom/components.py
```diff
@@ -47,98 +47,118 @@ def view_to_component(
...
-        """Render the view in an async hook to avoid blocking the main thread."""
-        # Render Check 1: Compatibility mode
-        if compatibility:
-            dotted_path = f"{view.__module__}.{view.__name__}"
-            dotted_path = dotted_path.replace("<", "").replace(">", "")
-            IDOM_VIEW_COMPONENT_IFRAMES[dotted_path] = ViewComponentIframe(
-                view, args, kwargs
-            )
+    @component
+    def new_component():
```
rmorshea, line 62: Perhaps a better way to do this would be to have a `make_view_component` function of the form:

```python
def make_view_component(view, compatibility=False, transforms=(), strict_parsing=True):
    @component
    def view_component(*args, request=None, **kwargs):
        ...
    return view_component
```

Where usage would look like:

```python
my_view_component = make_view_component(
    view,
    compatibility=True,
    transforms=[],
    strict_parsing=True,
)

@component
def example_usage():
    return my_view_component(*args, request=None, **kwargs)
```

This is convenient since you don't need to pass in the `view` every time and `args` and `kwargs` can be passed in normally rather than as a list and dict respectively.

Archmonger: But altering the args and kwargs based on an arbitrary `state` is a feature I'd like to retain.

rmorshea: Which args/kwargs are you refering to? What I propose would allow the args/kwargs passed to the view function to depend on state. By contrast, things like `compatibility`, `transforms`, and `strict_parsing` seem like they could remain static.

rmorshea: Ah, I didn't catch that compatibility was using in the component. Maybe the interface should be:

```python
my_view_component = make_view_component(view)

@component
def example_usage():
    return my_view_component(
        *args,
        request=None,
        compatibility=True,
        transforms=[],
        strict_parsing=True,
        **kwargs,
    )
```

However, like you said, this eats up a fair bit of the parameter namespace for the view function. One way to work around this would be to allow these settings to be passed in as an optional positional argument.

```python
my_view_component = make_view_component(view)

@component
def example_usage():
    return my_view_component(
        ViewOptions(
            request=None,
            compatibility=True,
            transforms=[],
            strict_parsing=True,
        ),
        *args,
        **kwargs,
    )
```

From a typing perspective, the arguments for the original `view` can be preserved using [`ParamSpec` and `Concatenate`](https://peps.python.org/pep-0612). Would be happy to contribute the typing part if it's unclear how that applies.

rmorshea: I'm also realizing that the layout [checks the identity](https://github.com/idom-team/idom/blob/main/src/idom/core/layout.py#L405) of the function that was used to render the component as one way to determine whether or not an element needs to be re-rendered. By re-defining the component on every render, this will confuse the layout into thinking this is a new type of component.

Archmonger: That interface would have the same issue of being unable to do early registration of `compatibility` components.

I think I just thought up a solution to component identity. Let me know if you like it.

Archmonger: I decided my situation is not a normal use-case so I refactored Conreq to have static URLs for several things that were previously using `compatibility`.

Do we still want to consider replacing `view_to_component` with `make_view_component`?

rmorshea: I think I prefer the `make_view_component` interface. It seems convenient to not need to pass all the other arguments every time.

rmorshea: Looking while looking into how this should be annotated, I've discovered there's some tradeoffs to consider.

### Option 1

The current `view_to_component` function can be typed in the following way:

```python
P = ParamSpec("P")

def view_to_component(
    view: Callable[P, HttpResponse],
    request: HttpRequest | None = None,
    compatibility: bool = True,
    transforms: Sequence[TransformFunc] = (),
    strict_parsing: bool = True,
    *args: P.args,
    **kwargs: P.kwargs,
) -> Component:
    ...
```

This has the advantage of providing accurate type hints for `*args, **kwargs` passed to `view` in IDEs. However, this comes at the cost of needing to pass in all the other arguments every time.

### Option 2

We can annotate a `make_view_component` function in the following way:

```python
P = ParamSpec("P")

def make_view_component(
    view: Callable[P, HttpResponse],
    compatibility: bool = True,
    transforms: Sequence[TransformFunc] = (),
    strict_parsing: bool = True,
) -> Callable[P, Component]:
    ...
```

This has the pros of adding accurate type hints in IDEs for the parameters of the `view` function as in option 1. It also has the benefit that potentially redundant arguments like `transforms` and `strict_parsing` do not need to be passed every time.

However, this has the con that, because there is no way to annotate that we can pass in an `HttpRequest` automatically in Python's type system. The user will have to manually pass an `HttpRequest` object to the returned component constructor:

```python
my_view_component = make_view_component(view)

@component
def example():
    return my_view_component(HttpRequest())
```

### Option 3

We could lastly annotate `make_view_component` but without providing hints for `*args, **kwargs` to the `view` function

```python
def make_view_component(
    view: Callable[..., HttpResponse],
    compatibility: bool = True,
    transforms: Sequence[TransformFunc] = (),
    strict_parsing: bool = True,
) -> Callable[..., Component]:
    ...
```

This has the cost of not providing type hints for the arguments of `view`, but it has the benefit that we aren't constrained by the type system and can automatically pass in the `request` object if not explicitely passed by the user

```python
my_view_component = make_view_component(view)

@component
def example():
    return my_view_component()
```

rmorshea: I think option 2 is my preference, but I'm also not sure how weird passing the request object would seem to Django users.

Archmonger: `HttpRequest.method = None` definitely breaks all class based views, and most function views. Personally, the currently API isn't that bad because [99% of the time](https://github.com/Archmonger/Conreq/blob/a0fccd3b242419f7f99829a7e9e5101e7e4bfd79/conreq/_core/server_settings/components.py#L20) the only arguments being passed in are `view` and _maybe_ also `compatibility`.

I personally am leaning towards option 1, and then shift to option 2 when a Python version that solves the type hinting constraints becomes Django's minimum version.

rmorshea: So, the annotations I suggested for option 1 don't work for other reasons. For example, in this case, MyPy has the following complaint:

```python
from typing import Callable, ParamSpec, Any

P = ParamSpec("P")

def view_to_component(
    func: Callable[P, None],
    x: int,
    *args: P.args,
    **kwargs: P.kwargs,
) -> Any:
    ...
    
# error: Missing positional argument "x" in call to "view_to_component"
view_to_component(lambda x: None, 1)
```

This is because MyPy doesn't realize that `x` in `view_to_component` will be passed to the lambda.

For this reason, I think we should give up on trying to type this correctly. At which point, we could just go with option 3. In the case where, 99% of the time you're only passing the view and something like `compatibility`, with `make_view_component` you end up passing 0 arguments to the constructor which seems like a nice benefit.

Archmonger: I think I found a happy middle ground between type hints and usability. Let me know what you think.

src/django_idom/components.py
```diff
@@ -83,62 +67,121 @@ async def async_renderer():
...
+    """
+
+    def decorator(view: Callable | View):
+        if not view:
+            raise ValueError("A view must be provided to `view_to_component`")
+
+        def wrapper(
+            request: HttpRequest | None = None,
+            args: Sequence | None = None,
+            kwargs: Dict | None = None,
```
rmorshea, line 138: Why not make this `*args, **kwargs`? Seems like a better user experience than passing a tuple and dict directly.

Archmonger: No way to type hint `request, *args, **kwargs` properly.

rmorshea: It's valid to annotate this as `request: HttpRequest | None = None, *args: Any, **kwargs: Any`

Archmonger: Yes but what about [line 113](https://github.com/Archmonger/django-idom/blob/ba93fcd5fbd8d2bd6c7022fb4d5c712165d165ea/src/django_idom/components.py#L113)?

rmorshea: It's ugly, but you'd annotate that using a protocol:

```python
class _ViewComponentConstructor(Protocol):
    def __call__(self, request: HttpRequest, *args: Any, **kwargs: Any) -> ComponentType:
        ...

def view_to_component(f: Callable, ...) -> _ViewComponentConstructor: ...
```

rmorshea (approved):

PR #101, opened 2022-10-06
Hook Cleanup

Archmonger:
## Description

This PR covers any oddities I discovered while integrating the new hooks with Conreq.

- [x] Allow `use_mutation` to have `refetch=None`, as the docs suggest is possible.
- [x] `use_origin` hook to return the browser's `location.origin`
- [x] Fix flaky `test_use_query_and_mutation` test (Key presses require at least 100ms between `keyup`/`keydown` on GH actions)
- [x] Logging for `use_mutation`/`use_query` errors (currently silent, which is very hard to debug)
- [x] Document `use_mutation`'s  `reset()`
- [x] Document `use_query` requires a `Model` or `QuerySet` return (due to enforcing no lazy queries)
- [x] Document interface on all APIs
- [x] Documentation misc cleanup and styling

## Checklist:


Archmonger:
@rmorshea I'm detecting something extremely fishy about how `use_query` operates.

If I manually disable `refetch` on the [two use_mutation calls](https://github.com/Archmonger/django-idom/blob/b73e9e23e404b026d7f4ec1afc80b0684c0dbbef/tests/test_app/components.py#L181-L195), somehow `items.data` performs another ORM call without IDOM ever re-calling [`get_items_query`](https://github.com/Archmonger/django-idom/blob/b73e9e23e404b026d7f4ec1afc80b0684c0dbbef/tests/test_app/components.py#L156-L157).

You can test this by removing the two `refetch` calls, then manually toggling the checkboxes in Chrome.

This might be indicative of a bigger problem with how we expect the Django models queries to operate.

But before I go down that rabbit hole, I'd like you to confirm my findings are not just a simple mistake.

Archmonger:
Did some more debugging and determined there is no issue.

`toggle_item_mutation` changes the value of `item.done` by reference, which allows visual updates without `refetch`.

Unrelated note, [`deferred_fields` are not equivalent](https://pandafy.me/deferred-fields-in-django) to lazy fields (all fields are lazy). So I am altering the fetch logic.

src/django_idom/hooks.py
```diff
@@ -24,13 +27,30 @@
...
 def use_location() -> Location:
-    """Get the current route as a string"""
+    """Get the current route as a `Location` object"""
     # TODO: Use the browser's current page, rather than the WS route
     scope = use_scope()
     search = scope["query_string"].decode()
     return Location(scope["path"], f"?{search}" if search else "")
 
 
+def use_origin() -> str | None:
```
rmorshea, line 37: I think we should limit what hooks we include to those that solve problems specific to the Django framework. This seems generally useful - something that could belong in a package that anyone using IDOM could take advantage of.

rmorshea (changes requested):
Since IDOM uses [Google's style guide](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings) for docstrings, let's stick to that for consistency.

I've also been wondering whether we should add a parameter to `use_query` which would remove the restriction that the return type be a `Model` or `QuerySet`. Perhaps we can just create an issue for now, but I suspect someone is going to have a performance issue that would best be resolved by limiting what fields are returned.

src/django_idom/hooks.py
Archmonger: The `use_origin` hook is specific to Django Channels. It's something I determined would be frequently used when creating SPAs with `django`+`idom`.

This is fairly easy to obtain when using Django HTML templates (http), but is very awkward to obtain when using IDOM (websockets) for rendering.

Archmonger:
Agreed that there should be an issue to modify `use_query` to remove restrictions on `QuerySet` and `Model`. Although the vast majority of Django users will use the built-in Django ORM, not all users do. [SQLAlchemy](https://www.sqlalchemy.org/) and [encode/orm](https://github.com/encode/orm) are infrequently used with Django on large-scale projects.

I don't believe any other ORM has the same sync/async context protection as Django. That issue is unique to Django due to how all queries are required to be lazy. Truthfully, I really wish Django core would just give us a `Model.lazy` toggle, or a `Model.execute` function so we can drop this awkwardness entirely.

src/django_idom/hooks.py
```diff
@@ -97,8 +128,19 @@ def execute_query() -> None:
...
 ) -> Mutation[_Params]:
+    """Hook to create, update, or delete Django ORM objects.
+
+    Args:
+        mutate: A callable that performs Django ORM create, update, or delete
+            functionality. If this function returns `False`, then your `refetch`
+            function will not be used.
+        refetch: A callable or sequence of callables that will be called if the
+            mutation succeeds. This is useful for refetching data after a mutation
+            has been performed.
```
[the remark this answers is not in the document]
Archmonger: From a technical perspective I agree. That description was me trying to find a balance on "what the user should know". I might have gone too sparse on the details here.

I think we'll need to scratch our heads on how to simplify the description of how things work.

rmorshea (approved):

PR #107, opened 2022-10-18
v2.0.1

Archmonger:
## Description

Forgot to pass through the `key=...` parameter on components.

## Checklist:

### DIFF
--- a/src/django_idom/components.py
+++ b/src/django_idom/components.py
@@ -11,7 +11,7 @@ from django.http import HttpRequest
 from django.urls import reverse
 from django.views import View
 from idom import component, hooks, html, utils
-from idom.types import ComponentType, VdomDict
+from idom.types import ComponentType, Key, VdomDict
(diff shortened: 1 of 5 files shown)


src/django_idom/components.py
```diff
@@ -164,31 +166,35 @@ def _django_css(static_path: str):
     return html.style(_cached_static_contents(static_path))
 
 
-def django_css(static_path: str):
+def django_css(static_path: str, key: Any | None = None):
```
rmorshea, line 169: key type is `str | None`

Archmonger: I thought keys could be int as well?

rmorshea: Ah, you're right. There's actually a type you can just import for this. `idom.types.Key`. So the type would be `Key | None`

rmorshea (approved):

PR #108, opened 2022-10-19
v2.1.0

Archmonger:
## Changelog

- Change type hint on `view_to_component` callable to have `request` argument be optional.
- More DRY way of async view rendering.
- Have docs demonstrate how to use Django-IDOM with `channels>=4.0.0`.
- Concatenate some examples on the `view_to_component` docs
- Add note to docs about potential information exposure via `view_to_component` when using `compatibility=True`.
- Clean up docs on running tests

## Checklist:

### DIFF
--- a/requirements/check-types.txt
+++ b/requirements/check-types.txt
@@ -1,2 +1,1 @@
 mypy
-django-stubs[compatible-mypy]
--- a/requirements/check-types.txt
+++ b/requirements/check-types.txt
@@ -1,1 +1,2 @@
 mypy
+django-stubs[compatible-mypy]
--- a/src/django_idom/components.py
+++ b/src/django_idom/components.py
@@ -20,7 +20,7 @@ from django_idom.utils import _generate_obj_name
 
 class _ViewComponentConstructor(Protocol):
     def __call__(
-        self, request: HttpRequest | None, *args: Any, **kwargs: Any
+        self, request: HttpRequest | None = None, *args: Any, **kwargs: Any
     ) -> ComponentType:
         ...
 
--- a/src/django_idom/components.py
+++ b/src/django_idom/components.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import json
 import os
 from inspect import iscoroutinefunction
-from typing import Any, Callable, Protocol, Sequence
+from typing import Any, Callable, Protocol, Sequence, overload
 
 from channels.db import database_sync_to_async
 from django.contrib.staticfiles.finders import find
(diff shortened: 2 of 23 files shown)


Archmonger:
I need some support on the return type hint for `view_to_component`

Since the decorator is polymorphic, I can't figure out how to annotate how this decorator can be used with or without parenthesis.

The current annotations do not work properly when parenthesis are used, such as [in our tests](https://github.com/Archmonger/django-idom/blob/0d05e6ed883952ff32f370220e6d4a7564191e5b/tests/test_app/components.py#L383).

rmorshea:
You'd handle that polymorphic behavior with [`@overload`](https://mypy.readthedocs.io/en/stable/more_types.html#function-overloading). Typing this correctly can tricky. Don't have time to give a full example now, but can help if needed.

Archmonger:
Support on that would be great.

You can either commit it directly to this branch, or we can push this PR forward and PR proper type hints later.

rmorshea:
Here's a sample of what the overload might look like:

```python
@overload
def view_to_component(view: Callable | View) -> _ViewComponentConstructor: ...

@overload
def view_to_component(view: None = ..., other_params: Any) -> Callable[[Callable], _ViewComponentConstructor]: ...

# final definition
def view_to_component(view: Callable | View | None = None, other_params: Any) -> _ViewComponentConstructor | Callable[[Callable], _ViewComponentConstructor]: ...
```

Archmonger:
Dang, overload is pretty neat.

I needed to do some modifications in order for the type hints to also work for use cases such as:

```python
_view_to_component_kwargs = view_to_component(views.view_to_component_kwargs, compatibility=True)
```

I believe I got something pretty solid in terms of type hints.

Archmonger:
@rmorshea not sure why my local MyPy doesn't generate warnings but GitHub actions does.

requirements/check-types.txt
```diff
@@ -1,2 +1 @@
 mypy
-django-stubs[compatible-mypy]
```
rmorshea, line 2: You removed this from the requirements file. I think this is causing the issue.

rmorshea: The underlying problem is that, without this, `View` is treated as `Any` and thus causes the signature overlap.

rmorshea (changes requested):
In general there seem to be a bunch of `type: ignore` comments that could potentially be resolved in a better way. Let's see if we can remove those. If not, could you comment with the reason?

For the ignore comments in our tests, I don't think those are necessary seeing as we don't actually type check our tests. If your IDE is showing type errors in those files, let's add an [exclude pattern](https://mypy.readthedocs.io/en/stable/config_file.html#confval-exclude) to our config.

src/django_idom/components.py
```diff
@@ -68,56 +68,83 @@ async def async_render():
...
                     }
-                )
+                )  # type: ignore
             )
             return
 
         # Render Check 2: Async function view
         elif iscoroutinefunction(view):
-            view_html = await view(request_obj, *args, **kwargs)
+            view_html = await view(request_obj, *_args, **_kwargs)  # type: ignore
```
rmorshea, line 77: This seems to be a problem because the type of `view` ought to be `type[View]` instead of a `View` instance.

src/django_idom/components.py
```diff
@@ -68,56 +68,83 @@ async def async_render():
...
+            )
 
         # Signal that the view has been rendered
         set_converted_view(
             utils.html_to_vdom(
                 view_html.content.decode("utf-8").strip(),
                 *transforms,
                 strict=strict_parsing,
-            )
+            )  # type: ignore
```
rmorshea, line 108: Let's `typing.cast` the initial value of converted view to `VdomDict` so we can remove this ignore comment.

Archmonger:
I think it's worthwhile to keep the ignores in `tests/*`. It actually helped out in letting me detect when the `@view_to_component` decorator type hints were broken.

rmorshea:
Perhaps we should run `mypy` on our tests then?

Archmonger:
Done

Archmonger:
Hopefully we can push this out today. I'd like to get the VTC security warning on the docs ASAP

rmorshea (changes requested):

rmorshea (approved):

rmorshea:
I think we can merge this, but I'm a little uncertain whether running MyPy against our test suite is the right move. I'm worried that the benefits of doing so don't outweigh the burden it might places on potential contributors. MyPy allows for per-module configuration, so perhaps, in a follow-up PR, we could tweak our MyPy configuration for our tests to be more lenient than the configuration we apply to our tests.

Archmonger:
I don't mind not running the test suite on the `tests/` folder via workflows. But I'd like to be able to verbosely see errors when actually typing the tests.

Increasing test leniency might reduce the API/typehint bugs I manually catch while writing test.

rmorshea:
MyPy is pretty configurable so I'm pretty sure we could strike a nice balance.

PR #117, opened 2023-01-02
v2.2.1: Fix recursive fetch depth for ManyToOneRel

### LINKED ISSUE #116: `use_query` ManyToMany fetching sometimes fails
### Current Situation

Fetching [subcategories in Conreq](https://github.com/Archmonger/Conreq/blob/3d10c7f4db84cd3db1d4dd98d70d521c69437aa3/conreq/_core/app_store/models.py#L119) does not automatically occur with `use_query`

### Proposed Actions

Determine why and push a fix.

Archmonger:
## Description

Recursive fetching was only ocurring on `ManyToManyField`. This PR makes the recursive call occur for both `ManyToOneRel` and `ManyToManyField`.

## Changelog
- fix #116
- Remove flaky `idom.web.export` test

## Checklist:


tests/test_app/components.py
```diff
@@ -44,15 +44,6 @@ def parameterized_component(x, y):
...
 
-victory = web.module_from_template("react", "victory-bar", fallback="...")
-VictoryBar = web.export(victory, "VictoryBar")
-
-
-@component
-def simple_bar_chart():
-    return html._(VictoryBar(), html.hr())
-
-
```
rmorshea, line 55: We can remove the use of `module_from_template`, however we do need something that tests whether or not web modules load properly. The most reliable way to do that would be to use `module_from_file` and write a super simple JS module. We [do this](https://github.com/idom-team/idom/blob/976028073fc3fba43870fdbeb3af96cad6e77ed1/tests/test_web/js_fixtures/simple-button.js) in IDOM core presently.

tests/test_app/components.py
```diff
@@ -44,6 +44,20 @@ def parameterized_component(x, y):
     )
 
 
+SimpleButtonModule = web.module_from_url(
```
rmorshea, line 47: We want this to specifically be `module_from_file` rather than from URL because we want to exercise web modules loaded from [this view](https://github.com/idom-team/django-idom/blob/main/src/django_idom/http/views.py#L12).

Archmonger: It has been switched.

Archmonger:
@rmorshea can we merge this?

Will push this version bump out then work on #118

rmorshea (approved):

PR #120, opened 2023-01-10
Database-backed component params

Archmonger:
## Description

Functional implementation of databased-backed component params

## Checklist:

-   [ ] Tests have been included for all bug fixes or added functionality.
-   [ ] The changelog has been updated with any significant changes, if necessary.
-   [ ] GitHub Issues which may be closed by this PR have been linked.


Archmonger:
@rmorshea This is fully functional but currently has no eviction/expiration strategy implemented. Let me know what you think.

rmorshea:
Will take a look tomorrow

src/django_idom/templatetags/idom.py
```diff
@@ -34,17 +35,23 @@ def component(dotted_path: str, **kwargs: Any):
...
-    _register_component(dotted_path)
-
+    component = _register_component(dotted_path)
+    uuid = uuid4().hex
     class_ = kwargs.pop("class", "")
-    json_kwargs = json.dumps(kwargs, separators=(",", ":"))
+
+    # Store the component's args/kwargs in the database if needed
+    # This will be fetched by the websocket consumer later
+    if func_has_params(component):
```
rmorshea, line 44: I don't think we should quietly ignore parameters that may not be consumed by the component. It seems better to let the error happen and allow the user to correct it. I understand this is a performance optimization, but I think we can worry about that if/when it turns out to be necessary to make things more efficient.

rmorshea:
Overall LGTM. Perhaps we could add a test which passes a parameter to a component that isn't JSON serializable?

src/django_idom/templatetags/idom.py
Archmonger: I [added in a check](https://github.com/idom-team/django-idom/pull/120/commits/e686894db9557ffbf8f30c0573b766f669f51c38) to catch this edge case.

I'd really like to avoid hitting the DB when possible, given how slow the Django ORM is.

rmorshea: If that's the case, I'd recommend using `inspect.Signature` [to test](https://docs.python.org/3/library/inspect.html#inspect.Signature.bind) whether or not the `kwargs` will work with the given function:

```python
kwargs = {...}
func = lambda: ...
sig = inspect.signature(func)
# check if kwargs work with func
sig.bind(kwargs)
```

If you check this when rendering the template, you an safely avoid hitting the DB if the func has no args/kwargs at runtime.

Archmonger: Implemented in https://github.com/idom-team/django-idom/pull/120/commits/4f4197513878cc8c8672d07579a8e71e6266628a

src/django_idom/websocket/consumer.py
```diff
@@ -64,7 +64,7 @@ async def _run_dispatch_loop(self):
             ),
             carrier=WebsocketConnection(self.close, self.disconnect, dotted_path),
         )
-        component_args: tuple[Any, ...] = tuple()
```
rmorshea: lowercase tuple is actually fine if you add `from __future__ import annotations`

Archmonger:
The reconnect variable has been changed to better suite the new behavior. Also, it is changed in preparation for https://github.com/idom-team/idom/issues/525

In the future, I hope to introduce a `IDOM_RECONNECT_INTERVAL` setting as well.

Archmonger:
Changelog in the PR's description has been updated.

Archmonger:
Lemme know when you can do the final review.

rmorshea:
this evening or tomorrow morning

docs/src/contribute/code.md
```diff
@@ -10,7 +10,7 @@ If you plan to make code changes to this repository, you will need to install th
 
 Once done, you should clone this repository:
 
-```bash
+```bash linenums="0"
```
rmorshea, line 13: Is there any way to create a macro or configure this to be the default. I feel like enforcing this as a consistent style will be tedious.

More of an observation than a requirement for merging this.

src/django_idom/hooks.py
```diff
@@ -65,17 +65,14 @@ def use_origin() -> str | None:
...
 
-def use_websocket() -> IdomWebsocket:
-    """Get the current IdomWebsocket object"""
-    websocket = use_context(WebsocketContext)
-    if websocket is None:
-        raise RuntimeError("No websocket. Are you running with a Django server?")
-    return websocket
+def use_connection() -> Connection:
+    """Get the current `Connection` object"""
+    return _use_connection()  # type: ignore
```
rmorshea, line 75: Should these hooks be deprecated in favor of the hooks from core?

docs/src/features/templatetag.md
```diff
@@ -44,7 +46,7 @@
     For this template tag, there are two reserved keyword arguments: `class` and `key`
 
     -   `class` allows you to apply a HTML class to the top-level component div. This is useful for styling purposes.
-    -   `key` allows you to force the component to use a [specific key value](https://idom-docs.herokuapp.com/docs/guides/understanding-idom/why-idom-needs-keys.html?highlight=key). You typically won't need to set this.
+    -   `key` allows you to force the component to use a [specific key value](https://idom-docs.herokuapp.com/docs/guides/understanding-idom/why-idom-needs-keys.html?highlight=key). Using `key` within a template tag is effectively useless.
```
rmorshea, line 49: Better link for keys: https://idom-docs.herokuapp.com/docs/guides/creating-interfaces/rendering-data/index.html#organizing-items-with-keys

src/django_idom/types.py
```diff
@@ -16,26 +17,40 @@
...
 
 _Result = TypeVar("_Result", bound=Union[Model, QuerySet[Any]])
 _Params = ParamSpec("_Params")
 _Data = TypeVar("_Data")
 
 
 @dataclass
-class IdomWebsocket:
-    """Websocket returned by the `use_websocket` hook."""
+class Connection:
```
rmorshea, line 34: You can subclass core's `Connection`:

```python
from idom.types import Connection as _Connection

Connection = _Connection[ComponentWebsocket]
```

rmorshea (commented):
Does this test the component parameter cleanup logic? If not, I think we should find a way to do that.

src/django_idom/types.py
Archmonger: Subclassing core's connection won't allow me to type hint `carrier` though.

src/django_idom/hooks.py
Archmonger: I tried that out initially, and it felt wrong. The documentation for how those specific hooks behave with Django-IDOM are going to be long in this repo. Feels weird to then have the user's import path be in core.

Plus `use_origin` wouldn't be in core, so that's yet another mismatch.

docs/src/contribute/code.md
Archmonger: That ironically is the default behavior. But at some point I realized we had way more code examples than bash examples, so I changed the default to enable line numbers to show.

I can change the default back, but then I'd need to manually write `linenums=1` on way more LOCs.

src/django_idom/types.py
Archmonger: Scratch that, I see how you set it up.

Archmonger:
I've added tests for the component parameter cleanup logic

Archmonger:
Let me know if this is mergeable

Archmonger:
@rmorshea This has security fixes, so I'd like to get this out soon if possible.

Archmonger:
I removed the version bump from this PR.

I will create a follow-up PR that bumps IDOM to `1.0.0` and Django-IDOM to `3.0.0`.

rmorshea:
Will review tomorrow.

docs/src/features/settings.md
```diff
@@ -9,16 +9,17 @@
...
     CACHES = {
     "idom": {"BACKEND": ...},
     }
 
-    # Maximum seconds between two reconnection attempts that would cause the client give up.
-    # 0 will disable reconnection.
-    IDOM_WS_MAX_RECONNECT_TIMEOUT = 604800
+    # Maximum seconds between reconnection attempts before giving up.
+    # Use `0` to prevent component reconnection.
+    IDOM_RECONNECT_MAX = 259200
```
rmorshea, line 20: To clarify, IDOM never actually give up trying to reconnect at the moment. The max reconnect is the max time between attempts.

src/django_idom/apps.py
```diff
@@ -1,11 +1,21 @@
...
     def ready(self):
         # Populate the IDOM component registry when Django is ready
-        ComponentPreloader().register_all()
+        ComponentPreloader().run()
+
+        # Delete expired database entries
+        # Suppress exceptions to avoid issues with `manage.py` commands such as
+        # `test`, `migrate`, `makemigrations`, or any custom user created commands
+        # where the database may not be preconfigured.
+        with contextlib.suppress(OperationalError):
```
rmorshea, line 20: Should we log the exception? Probably with debug or info level.

src/django_idom/templatetags/idom.py
```diff
@@ -34,17 +35,29 @@ def component(dotted_path: str, **kwargs: Any):
...
     return {
         "class": class_,
         "idom_websocket_url": IDOM_WEBSOCKET_URL,
         "idom_web_modules_url": IDOM_WEB_MODULES_URL,
-        "idom_ws_max_reconnect_timeout": IDOM_WS_MAX_RECONNECT_TIMEOUT,
-        "idom_mount_uuid": uuid4().hex,
-        "idom_component_id": dotted_path,
-        "idom_component_params": urlencode({"kwargs": json_kwargs}),
+        "idom_reconnect_max": IDOM_RECONNECT_MAX,
+        "idom_mount_uuid": uuid,
```
rmorshea, line 61: Just as a final check, could an attacker leverage this UUID in any way? For example, if credentials are passed via this mechanism, could an attacker attempt to re-use this UUID somehow?

rmorshea (commented):
Generally approve, just few minor nits/comments

src/django_idom/templatetags/idom.py
Archmonger: Yes, but the same vulnerability exists for session cookies and CSRF tokens as well. The defense mechanism comes from the fact that the uuid has enough entropy to avoid random guessing. The only ways of abuse would be brute force or man-in-the-middle attacks, both of which can be mitigated by proper webserver/firewall config.

docs/src/features/settings.md
Archmonger: It's a bit of a preemptive rename and description change.

See https://github.com/idom-team/django-idom/pull/120#issuecomment-1378170208

src/django_idom/apps.py
Archmonger: Sure will do

src/django_idom/templatetags/idom.py
Archmonger: In the future we could potentially make this more secure through public/private key pairs (similar to SSL) but that gets into the territory of encryption, which should be handled in a separate PR.

rmorshea (approved):
I think everything looks good to me.

PR #131, opened 2023-03-01
Rename `django-idom` to `reactpy-django`

Archmonger:
## Description

Performs any changes needed to rename from `idom` to `reactpy`.

## Checklist:

-   [ ] The changelog has been updated with any significant changes, if necessary.


Archmonger:
Couple of things are needed for these tests to pass

1) `reactpy` 1.0.0 release for parity to `idom`
   - Why did we revert versions when renaming the package...?
2) `reactpy-router` stub repo
3) `django-reactpy` stub repo

rmorshea:
I guess I figured that if we wanted to release a 1.0 version of ReactPy we'd need to remove the existing tag. Releasing a 0.45.0 version was mostly a result of me rushing to get a release out after I cleared out the previous releases.

rmorshea:
Let's go with beginning all names with `reactpy-` for consistency. Also, why do we need stub repos? Can't we just rehome this repo?

rmorshea:
Ok I see. I think we can merge despite the docs build failing. We'll just rehome the repo and then rerun the tests.

Archmonger:
Can we re-release ReactPy as 1.0.0?

rmorshea:
[`@reactpy/client`](https://www.npmjs.com/package/@reactpy/client) has been released. You should be able to use this without any changes. The next release of the client (assuming there's nothing wrong with v0.1.0) will be the typescript rewrite.

Archmonger:
The version number regressions of the JS and Python packages are really unnatural.

rmorshea:
The zero version on the Python package is temporary. I'm working on publishing a 1.0 release momentarily.

rmorshea:
v1.0 of reactpy Python lib had been released.

rmorshea:
@Archmonger, because GitHub aggressively strips all style from inline HTML there is no way to copy-paste a common markdown header to all our repos with a simple "ReactPy" to "ReactPy <Whatever>" title change. From what I can tell, each repo will need its own landscape logo file with the modified title.

rmorshea:
The only alternative I can think of would be to do something like:

```
# <img src="https://raw.githubusercontent.com/reactive-python/reactpy/main/branding/svg/reactpy-logo-square.svg" alt="ReactPy Logo" style="height: 2rem; vertical-align: text-bottom" /> ReactPy
```

image

Since for some reason `img` elements seem to allow inline style. The current logo doesn't really work well at this scale, but perhaps the best option for right now.

Archmonger:
By landscape, do you mean center justified?

rmorshea:
By landscape I mean [this style of logo](https://github.com/reactive-python/reactpy/blob/reusable-readme-title/branding/svg/reactpy-logo-landscape.svg).

Archmonger:
I've done something similar using the `align` image property in the past.

```md
# ReactPy <img src="https://raw.githubusercontent.com/reactive-python/reactpy/main/branding/svg/reactpy-logo-square.svg" align="left" height="40"/>
```

# ReactPy

rmorshea:
Let's go with the logo setup you have there. After that I think this LGTM.

Archmonger:
The tagline didn't really have a good place with the new logo design, so I opted to remove it. I think we should just keep that text in the github sidebar moving forward.

Take a look at the design here: https://github.com/Archmonger/django-idom/tree/django-reactpy

rmorshea:
do you think we need the backends table here?

Archmonger:
Not from a technical standpoint, but it's free advertising that doesn't feel like advertising.

The more click-throughs we get, the more we can snowball our overall popularity.

rmorshea:
I was planning to rehome the repo right after merging so perhaps we could leave the urls as is.

Archmonger:
It's very little extra effort to do a follow up PR to change URLs.

We're already seeing some instances of confusion on the IDOM core docs.

rmorshea (approved):
I'm good with things as they are so approving. feel free to make those url changes though.

rmorshea:
Good to merge?

Archmonger:
That last commit revealed some holes in my link checking workflow.

Since the `README` and `CHANGELOG` aren't in the docs directory they aren't getting checked.

Also one more thing I want to revert is maintaining the `idom` name in the legacy changelog versions.

rmorshea:
Ok, just at me when it's ready and I'll merge/rehome.

Archmonger:
Since I didn't revert any `reactpy-django` references in the docs, I think we should repo transfer before merging this PR to prevent broken docs links.

Also, our logo solution doesn't play well with `mkdocs` so I might have to do some changes to make that work.

rmorshea:
LGTM!

PR #154, opened 2023-06-25
Add error checking to `component` template tag

Archmonger:

## Description

There are some realistic edge cases where template rendering can occur before the `reactpy_django` database table exists (#152). This scenario is not common, but can exist if using HTML postprocessors.

- Make it so that `component` template tag render failures don't cause catastrophic failure.
- `component` template tag render failures now output a visual indicator if `REACTPY_DEBUG_MODE` is enabled.
- Render `<hr>` via raw HTML within the tests so that component render failures don't break formatting
- Fix typo on the docs
- Ensure `generate_object_name` always returns a string

## Checklist:


src/reactpy_django/utils.py
```diff
@@ -291,15 +294,15 @@ def django_query_postprocessor(
...
 
-def func_has_params(func: Callable) -> bool:
-    """Checks if a function has any args or kwarg parameters."""
+def func_has_args(func: Callable) -> bool:
+    """Checks if a function has any args or kwarg."""
     signature = inspect.signature(func)
 
     # Check if the function has any args/kwargs
-    return str(signature) != "()" or True
+    return str(signature) != "()"
```
rmorshea: If this isn't used in multiple places perhaps we can just embed the logic where it's currently being used.

Archmonger: This comment actually made me realize I forgot to add in a check for this in the template tag.

But yes, it should be used in at minimum two locations. The consumer and the template tag.

I will add a regression test.

rmorshea (approved):

PR #170, opened 2023-08-09
v3.3.2

Archmonger:

### Added

-   ReactPy Websocket will now decode messages via `orjson` resulting in an ~6% overall performance boost.
-   Built-in `asyncio` event loops are now patched via `nest_asyncio`, resulting in an ~10% overall performance boost on Windows. This has no performance impact if you are running your webserver with `uvloop`.

### Fixed

-   Fix bug where `REACTPY_WEBSOCKET_URL` always generates a warning if unset.
-   Fixed bug where `assert f is self._write_fut` would be raised within `uvicorn` on Windows when `REACTPY_BACKHAUL_THREAD = True`.
-   Fixed bug where rendering behavior would be jittery with `daphne` on Windows when `REACTPY_BACKHAUL_THREAD = True`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,10 @@ Using the following categories, list your changes in this order:
 
 -   Changed implementation of `REACTPY_BACKHAUL_THREAD` to attempt increased performance compatibility.
 
+### Fixed
+
+-   Fix bug where `REACTPY_WEBSOCKET_URL` always generates a warning if unset.
+
 ## [3.3.1] - 2023-08-08
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,13 +41,13 @@ Using the following categories, list your changes in this order:
 ### Added
 
 -   ReactPy Websocket will now decode messages via `orjson` resulting in an ~6% overall performance boost.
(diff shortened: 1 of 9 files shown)


Archmonger:
@rmorshea I'm going to merge this as a temporary hotfix, but I'm honestly kind of shocked using `nest_asyncio` fixes both `assert f is self._write_fut` exceptions on Uvicorn and also jittery rendering behavior Daphne.

My best guess is the way that the Windows `ProactorEventLoop` enqueues tasks is inherently not thread safe, and maybe `nest_asyncio` [serializes this queue](https://github.com/erdewit/nest_asyncio/blob/2867d74504c19d0e77a4bab1ced957bafa772d32/nest_asyncio.py#L76-L99)? I'm even more shocked that this results in a overall performance improvement. 

There's still one open bug: [Some](https://github.com/reactive-python/reactpy-django/discussions/166#discussioncomment-6660262) Windows environments have degraded performance when running via backhaul thread. I've actually experienced this [in the past](https://github.com/reactive-python/reactpy-django/pull/23#issuecomment-956584980), but I've been unable to replicate it.

Will need to do a deeper dive into this in the future. Given that the issue existed for me before but is now gone, this is likely somehow related to CPython JIT optimizations.

PR #186, opened 2023-09-15
SEO Compatible Rendering

Archmonger:

## Description

- SEO compatible rendering
    - Prerenders the initial component via the template tag using `vdom_to_html`, then loads the actual component in the background within a `hidden` div. When loaded, the prerender is replaced with the actual render.
    - `settings.py:REACTPY_PRERENDER` can be set to `True` to enable this behavior by default
    - Enable it on individual components via the template tag: `{% component "..." prerender="True" %}`
- Docs styling, verbiage, and formatting tweaks
- Rename undocumented utility function `reactpy_django.utils.ComponentPreloader` to `reactpy_django.utils.RootComponentFinder`.
- Fix JavaScript being via `pip install -e .` on Windows.
- Update PyPi package metadata
- Update pull request template

## Checklist:


Archmonger:
@rmorshea BTW after a lot of testing, the most rational approach for this was to prerender in a separate div, then replace it with the actual render when possible. The prerendering step is performed entirely by the template tag.

Saves us from rehydration headaches, and is the most compatible solution across multiple backends.

PR #190, opened 2023-09-19
`UserModel` related hooks, decorators, and settings 

Archmonger:
## Description

-   New Django `User` related features!
    -   `reactpy_django.hooks.use_user` can be used to access the current user.
    -   `reactpy_django.hooks.use_user_data` provides a simplified interface for storing user key-value data.
    -   `reactpy_django.decorators.user_passes_test` is inspired by Django's [`user_passes_test`](http://docs.djangoproject.com/en/dev/topics/auth/default/#django.contrib.auth.decorators.user_passes_test) decorator, but works with ReactPy components.
    -   `settings.py:REACTPY_AUTO_RELOGIN` will cause component WebSocket connections to automatically [re-login](https://channels.readthedocs.io/en/latest/topics/authentication.html#how-to-log-a-user-in-out) users that are already authenticated. This is useful to continuously update `last_login` timestamps and refresh the [Django login session](https://docs.djangoproject.com/en/dev/topics/http/sessions/).

## Checklist:

-   [x] Tests have been developed for bug fixes or new functionality.
-   [x] The changelog has been updated, if necessary.
-   [x] Documentation has been updated, if necessary.
-   [x] GitHub Issues closed by this PR have been linked.

<sub>By submitting this pull request you agree that all contributions comply with this project's open source license(s).</sub>


Archmonger:
@rmorshea I need your opinion on `use_user_data` interface I'm constructing in this PR. To save you some time, here's the short and sweet summary:

`use_user_data` allows you to store or retrieve data specific to the connection's `User` object. This `user_data` is persistently stored within the database and can be accessed like such:

```python
from reactpy import component, html
from reactpy_django.hooks import use_user_data

@component
def my_component():
    user_data, set_user_data = use_user_data()

    async def on_submit(event):
        if event["key"] == "Enter":
            key = str(len(user_data.current))
            merged_data = user_data.current | {key: event["target"]["value"]}
            set_user_data(merged_data)

    return html.div(
        html.div(f"Data: {user_data.current}"),
        html.div(f"Loading: {user_data.loading | set_user_data.loading}"),
        html.div(f"Error(s): {user_data.error} {set_user_data.error}"),
        html.input({"on_key_press": on_submit}),
    )
```

---

To get the conversation started, here's a few questions/concerns I have:

1. Right now `user_data.current` returns a dict... Should we instead replace this with some kind of manager API?
2. Does splitting the interface into `user_data, set_user_data` feel appropriate?
3. Is the `set_user_data` interface clear and logical? Specifically the fact that it has `loading`, `reset`, and `error` properties (identical to `use_mutation`).
4. Is the `user_data` interface clear and logical? Specifically the fact that it has `current`, `loading`, `refetch`, and `error` properties. This is slightly different than the equivalent `use_query` interface since I wanted to rename the `data` attribute to `current`.

rmorshea:
I'll try and take a look early next week.

Archmonger:
Would you have time to look at the interface proposed in the [above comment](https://github.com/reactive-python/reactpy-django/pull/190#issuecomment-1853341375) this weekend?

This PR is fairly bulky and has a lot of Django-only jargon, so a PR review won't be needed.

rmorshea:
Sure. I'll make time today or tomorrow.

rmorshea:
My take is that the marginal UX improvements made by creating special wrappers for the query and mutation objects may be outweighed by requiring the user to learn something new. It seems easier to explain to the user that `use_user_data` returns a tuple with a query and a mutation.

To address some of the usability concerns in a slightly simpler way, we could consider adding a `__call__` method to mutations and returning a named tuple so usage would be

```python
user_data = use_user_data()

# access data
user_data.query.data

def some_callback():
    new_user_data = user_data.query.data | {...}
    user_data.mutation(new_user_data)
```

Archmonger:
Agreed, will implement soon.

PR #199, opened 2023-10-09
Add Jinja support

### LINKED ISSUE #198: Support Jinja2 templates
### Current Situation

All current tests are made with Django Template Tags in mind.

I currently don't know if our existing template tag works on Jinja2. It _probably_ doesn't?

### Proposed Actions

Support Jinja2 and add tests for this configuration.

ShaheedHaque:
## Description

To reproduce the results to date requires the following example changes on the Django side...

1. Configure template files ending in ".jinja" to be processed via Jinja:
    ```python
        TEMPLATES = [
        {
            "BACKEND": "django_jinja.backend.Jinja2",  # Jinja backend
            'DIRS': [os.path.join(BASE_DIR, 'templates')],
            "APP_DIRS": True,
            "OPTIONS": {
                'context_processors': [
                    'django.contrib.auth.context_processors.auth',
                    'django.contrib.messages.context_processors.messages',
                    ...
                ],
                "match_extension": (".jinja"),  # suffix for Jinja templates
                "app_dirname": "templates",
    
                # Can be set to "jinja2.Undefined" or any other subclass.
                "undefined": None,
    
                "newstyle_gettext": True,
                "tests": {
                    # "mytest": "path.to.my.test",
                },
                "filters": {
                    # "myfilter": "path.to.my.filter",
                },
                "globals": {
                    # "myglobal": "path.to.my.globalfunc",
                },
                "constants": {
                    # "foo": "bar",
                },
                "extensions": [
                    "jinja2.ext.do",
                    "jinja2.ext.loopcontrols",
                    "jinja2.ext.i18n",
                    "django_jinja.builtins.extensions.CsrfExtension",
                    "django_jinja.builtins.extensions.CacheExtension",
                    "django_jinja.builtins.extensions.TimezoneExtension",
                    "django_jinja.builtins.extensions.UrlsExtension",
                    "django_jinja.builtins.extensions.StaticFilesExtension",
                    "django_jinja.builtins.extensions.DjangoFiltersExtension",
                    "reactpy_django.templatetags.jinja.ReactPyExtension",    # new extension
                ],
                "bytecode_cache": {
                    "name": "default",
                    "backend": "django_jinja.cache.BytecodeCache",
                    "enabled": False,
                },
                "autoescape": True,
                "auto_reload": DEBUG,
                "translation_engine": "django.utils.translation",
            },
        },
        {
            'BACKEND': 'django.template.backends.django.DjangoTemplates', # Django templates.

    ```
2. Add the new app:
    ```python
    INSTALLED_APPS = [
         ...
        "reactpy_django",
     ]

    ```
3. File of components in `components.py`:
    ```python
    from reactpy import component, html

    @component
    def hello_world(recipient: str):
        return html.h1(f"Hello {recipient}!")
    ```
4. Test view in `views/react.py`:
    ```python
    from django.shortcuts import render

    def jinja_view(request):
        return render(request, "my-template.html.jinja")
    ```
6. Template for the view `templates/my-template.html.jinja`:
    ```html
    <!DOCTYPE html>
    <html>
    <body>
    {{ component("paiyroll.components.hello_world", recipient="World") }}
    </body>
    </html>
    ```
8. `project/asgi.py`
    ```python
    from reactpy_django import REACTPY_WEBSOCKET_ROUTE
 
    application = ProtocolTypeRouter({
      'http': get_asgi_application(),
      'websocket': AuthMiddlewareStack(
          URLRouter(
            ... + [REACTPY_WEBSOCKET_ROUTE]
         )
    ``` 

9. And finally `project/urls.py`:
    ```python
    from ...views import react
 
    urlpatterns = [
         ...
         path("reactpy/", include("reactpy_django.http.urls")),
         path("react/", react.jinja_view),
     ]
     ```
Once Django is restarted, navigating to the view, you should see
    ```
    Hello World!
    ```

## Checklist:

-   [ ] Tests have been developed for bug fixes or new functionality.
-   [ ] The changelog has been updated, if necessary.
-   [ ] Documentation has been updated, if necessary.
-   [ ] GitHub Issues closed by this PR have been linked.

<sub>By submitting this pull request you agree that all contributions comply with this project's open source license(s).</sub>


ShaheedHaque:
OK, this now render the "Hello World!" correctly.

Feedback requested before thinking about tests/docs etc.

src/reactpy_django/templatetags/jinja.py
```diff
@@ -0,0 +1,67 @@
...
+"""
+import typing as t
+
+from django.template import RequestContext, loader
+from jinja2 import pass_context
+from jinja2.ext import Extension
+from jinja2.runtime import Context, Undefined
+
+from .reactpy import component as djt_component
+from .. import config
```
Archmonger, line 13: No relative imports

src/reactpy_django/templatetags/jinja.py
```diff
@@ -0,0 +1,67 @@
...
+Jinja support.
+"""
+import typing as t
+
+from django.template import RequestContext, loader
+from jinja2 import pass_context
+from jinja2.ext import Extension
+from jinja2.runtime import Context, Undefined
+
+from .reactpy import component as djt_component
```
Archmonger, line 12: Use `import reactpy` instead, then call the component as `reactpy.component`

src/reactpy_django/templatetags/jinja.py
```diff
@@ -0,0 +1,67 @@
...
+                the host will be automatically configured. \
+                Example values include: `localhost:8000`, `example.com`, `example.com/subdir`
+            prerender: Configures whether to pre-render this component, which \
+                enables SEO compatibility and reduces perceived latency.
+            **kwargs: The keyword arguments to provide to the component.
+
+        Returns:
+            Whatever the components returns.
+        """
+        djt_context = RequestContext(__context.parent['request'], autoescape=__context.eval_ctx.autoescape)
```
Archmonger, line 61: Don't use acronyms when possible, to help future contributors. This variable should be named `request_context`

Archmonger (commented):
A few comments since I took a early peak at the draft.

Archmonger:
I don't think we should be storing components within `environment.globals["component"]`. That is both unnecessary memory overhead, and won't be multi-threading/multi-processing safe.

I would suggest the following interface instead:

`{{ component("dotted.path.to.component", args=[variable]) }}`

Also I know that Django template tags can be treated like normal functions, so we might want to call it directly within this Jinja tag if possible. That would save us the burden of keeping parity between their interfaces.

src/reactpy_django/templatetags/jinja.py
```diff
@@ -0,0 +1,67 @@
+# Copyright © 2023 Innovatie Ltd. All rights reserved.
+"""
+Jinja support.
+"""
+import typing as t
```
Archmonger, line 5: Import whatever you need `from typing import ...` instead of abstract import renaming.

src/reactpy_django/templatetags/jinja.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class ReactPyExtension(Extension):
+    """
+    Jinja has more expressive power than core Django's templates, and can
+    directly handle expansions such as:
+
+        {{ component(*args, **kwargs) }}
+    """
+    DJT_TEMPLATE = 'reactpy/component.html'
```
Archmonger, line 23: No point as storing this as a class attribute if it's only used once

src/reactpy_django/templatetags/jinja.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def __init__(self, environment):
+        super().__init__(environment)
+        #
+        # All we need is to add global "component" to the environment.
+        #
+        environment.globals["component"] = self._jinja_component
+
+    @pass_context
+    def _jinja_component(self, __context: Context, dotted_path: str, *args: t.Any, host: str | None = None,
```
Archmonger, line 37: May as well just accept `_jinja_component(self, __context: Context, *args, **kwargs)`.

That is, unless there's some tools out there for Jinja auto-complete within templates.

On that note, this method should be renamed to `def template_tag( ... )`.

src/reactpy_django/templatetags/jinja.py
```diff
@@ -0,0 +1,67 @@
...
+                Example values include: `localhost:8000`, `example.com`, `example.com/subdir`
+            prerender: Configures whether to pre-render this component, which \
+                enables SEO compatibility and reduces perceived latency.
+            **kwargs: The keyword arguments to provide to the component.
+
+        Returns:
+            Whatever the components returns.
+        """
+        djt_context = RequestContext(__context.parent['request'], autoescape=__context.eval_ctx.autoescape)
+        context = djt_component(djt_context, dotted_path, *args, host=host, prerender=prerender, **kwargs)
```
Archmonger, line 62: Rename to `django_template_tag = ...`

src/reactpy_django/templatetags/jinja.py
```diff
@@ -0,0 +1,67 @@
...
+        #
+        # TODO: can this be usefully cached?
+        #
+        result = loader.render_to_string(self.DJT_TEMPLATE, context, __context.parent['request'])
+        return result
```
Archmonger, line 67: Don't assign variables to something that is directly returned.

src/reactpy_django/templatetags/jinja.py
ShaheedHaque: This is really a reminder to pull this out into somewhere as thew value is required to be the same as used in the decorator for templatetags.component.

src/reactpy_django/templatetags/jinja.py
ShaheedHaque: Not sure what you mean, see updated code?

src/reactpy_django/templatetags/jinja.py
```diff
@@ -0,0 +1,63 @@
...
+    def __init__(self, environment):
+        super().__init__(environment)
+        #
+        # All we need is to add global "component" to the environment.
+        #
+        environment.globals["component"] = self.template_tag
+
+    @pass_context
+    def template_tag(self, jinja_context: Context, dotted_path: str, *args, host: str | None = None,
+                     prerender: str = str(config.REACTPY_PRERENDER), **kwargs) -> str:
```
Archmonger, line 35: In order to maintain interface parity more easily, the params for this can be changed to

`(self, jinja_context: Context, *args, **kwargs)`

ShaheedHaque: Yes, but then the structural similarity with the native code would be lost, and some annoying/ugly unpacking would be required on line 59. Surely that is worth avoiding?

Archmonger: The alternative is I would need to start manually maintaining interface parity between Jinja and Django every time I add a new template tag feature, which is a royal pain.

ShaheedHaque: Fair enough. One last thought...since `dotted_path` is required, how would you feel about:
```
def template_tag(self, jinja_context: Context, dotted_path: str, *args, **kwargs)
```

Archmonger: Yeah that works.

src/reactpy_django/templatetags/jinja.py
ShaheedHaque: @Archmonger I think this is the last remaining unresolved item. Any thoughts on sharing this value, or shall I just go ahead and hard code it?

Archmonger: Let's make it a global within the Django template tag module for now.

ShaheedHaque:
Resolved last outstanding comment. And ran black.

If we are to proceed, I'd propose to squash the current commits before thinking about tests/packaging/docs etc. (FWIW, in that regard, I assume the appropriate packaging would be as something like "reactpy_django[jinja]"...but I'm not currently sure how to go about implementing that).

Archmonger:
Optional dependecies can be defined within the package definition in [setup.py](https://github.com/reactive-python/reactpy-django/blob/6f79c4cf7ed8b0b4cd048817aa90ab12997caabc/setup.py#L47-L81). For example:

```python
{
  "extras_require": {
    "encryption": ["cryptography", "pycryptodome"],
  },
  ...
}
```

src/reactpy_django/templatetags/jinja.py
```diff
@@ -0,0 +1,66 @@
...
+from jinja2 import pass_context
+from jinja2.ext import Extension
+from jinja2.runtime import Context
+from reactpy_django import config
+from reactpy_django.templatetags.reactpy import component
+
+#
+# Point to our non-Django analogue.
+#
+DJT_TEMPLATE = "reactpy/component.html"
```
Archmonger, line 15: This does not solve the code duplication issue.

This should be stored within `reactpy_django.templatetags.reactpy` and imported into this file.

References to `"reactpy/component.html"` within `reactpy_django.templatetags.reactpy` will need to be replaced with this global.

src/reactpy_django/templatetags/jinja.py
```diff
@@ -0,0 +1,66 @@
+# Copyright © 2023 Innovatie Ltd. All rights reserved.
```
Archmonger, line 1: All code within this repository must be encompassed by our repository's [copyright licensing](https://github.com/reactive-python/reactpy-django/blob/main/LICENSE.md), as described in the pull request template.

Having your company retain copyright for this one file will not be acceptable. This would mean that legally we would have no right to use, copy, distribute, or modify your contributions.

src/reactpy_django/templatetags/jinja.py
ShaheedHaque: Ah, I misunderstood what you wanted. Getting rid of the duplication was my point too.

ShaheedHaque:
> Optional dependencies can be defined within the package in [setup.py](https://github.com/reactive-python/reactpy-django/blob/6f79c4cf7ed8b0b4cd048817aa90ab12997caabc/setup.py#L47-L81). For example:
> 
> ```python
> {
>   "extras_require": {
>     "encryption": ["cryptography", "pycryptodome"],
>   },
>   ...
> }
> ```

Thanks, I'm more or less familiar with this bit. I was more thinking about the new file I have added (and any supporting test file), and whether that could/should be omitted unless "[jinja] was specified?

Archmonger:
I have some changes in one of my PRs that will simplify test configuration
https://github.com/reactive-python/reactpy-django/pull/190

TLDR: Our test suite can now run multiple different Django settings.py files.

Archmonger:
My PR has been merged. All settings files [within the test app](https://github.com/reactive-python/reactpy-django/tree/main/tests/test_app) will now get the test suite automatically run against them.

There shouldn't be anything blocking this PR anymore, so you'll need to do the following:

1. Add your dependencies to [test-env.txt](https://github.com/reactive-python/reactpy-django/blob/main/requirements/test-env.txt)
2. Add the needed dependencies as [optional extras](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies) using the `[jinja]` keyword.
3. Cop [settings_single_db.py](https://github.com/reactive-python/reactpy-django/blob/main/tests/test_app/settings_single_db.py) and modifying it to fit your test case.
4. Write some [conditional tests](https://stackoverflow.com/questions/38442897/how-do-i-disable-a-test-using-pytest) within the [test folder](https://github.com/reactive-python/reactpy-django/tree/main/tests/test_app/tests)
5. Write a new page of documentation based on [the current installation page](https://github.com/reactive-python/reactpy-django/blob/main/docs/src/learn/add-reactpy-to-a-django-project.md).

ShaheedHaque:
Noted. I've a full plate right now but have this on my radar.

